PR #145, opened 2020-10-13
Block finality status - Closes #39

### LINKED ISSUE #39: Block finality status
### Description

The Service Blockchain API returns number of confirmations and does not associate it with any kind of information whether the block reached finality or not. 

This was previously up to the UI client based on the number of confirmations - see this [snippet](https://github.com/LiskHQ/lisk-explorer/blob/v2.3.0-rc.1/src/shared/transaction-item/transaction-item.html#L73).

Make sure that there is a reliable way to return that information from the API.

### Motivation

With the implementation of the BFT protocol in the core product the way a block is considered final has changed. That means that the arbitrarily chosen number of confirmations (101) is essentially incorrect and can be misleading.

### Acceptance Criteria

- Lisk Service API returns a transaction status based on the information provided by the research team.

### Additional Information

- [LIP: BFT consensus protocol](https://research.lisk.io/t/introduce-bft-consensus-protocol/139)

nagdahimanshu:
### What was the problem?
This PR resolves #39 

### How was it solved?

- [x] isFinal added to the Block Schema
- [x] Lisk Service API returns a Block status isFinal

### How was it tested?
Local + Jenkins


services/core/shared/core/compat/sdk_v3/blocks.js
```diff
@@ -0,0 +1,79 @@
...
+
+	const finalHeight = getFinalizedHeight();
+	const data = blocks.data.map((block) => {
+		if (block.height <= finalHeight) {
+			block.isFinal = true;
+		} else {
+			block.isFinal = false;
+		}
+		return block;
+	});
```
sameersubudhi, line 59:
```suggestion
	const data = blocks.data
		.map(block => block.isFinal = (block.height <= finalHeight));
```

services/core/events/blockchain.js
```diff
@@ -80,4 +80,17 @@ module.exports = [
...
 	},
+	{
+		name: 'update.height_finalized',
+		description: 'Keep the block finality height up-to-date',
+		controller: callback => {
+			coreSocket.socket.on('blocks/change', async () => {
+				let restData = null;
+				logger.debug('Returning latest heightFinalized to the socket.io client...');
+				if (core.updateFinalizedHeight) restData = await core.updateFinalizedHeight();
+				// restData = await core.updateFinalizedHeight();
```
sameersubudhi, line 91:
```suggestion
```

sameersubudhi (approved):
Except for the suggestions, LGTM overall. 👍

services/core/shared/core/compat/sdk_v3/blocks.js
nagdahimanshu: Done!

MichalTuleja (approved):

PR #149, opened 2020-10-15
Data transformer empty object not removed - Closes #143

### LINKED ISSUE #143: Data transformer: empty object not removed
### Actual behavior

Empty object on the top level is not removed from data structure.

### Expected behavior

Empty object on the top level is removed from data structure.

### Steps to reproduce

Call `/accounts` on a regular account, empty delegate property is visible.

sameersubudhi:
### What was the problem?
This PR resolves #143 

### How was it solved?
- [x] Remove any empty property present in the response object
- [x] Perform the check recursively

### How was it tested?
Local + Jenkins

### DIFF
--- a/tests/integration_testnet/api/rpc/accounts.test.js
+++ b/tests/integration_testnet/api/rpc/accounts.test.js
@@ -126,7 +126,7 @@ describe('Method get.accounts', () => {
 			expect(result.data[0].delegate).toMap(delegateSchema);
 		});
 
-		it('returns delegate data by delegate name', async () => {
+		xit('returns delegate data by delegate name', async () => {
 			const { result } = await getAccounts({ username: accounts.delegate.username });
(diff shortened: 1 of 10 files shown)


tests/integration_testnet/api/rpc/accounts.test.js
```diff
@@ -127,7 +127,7 @@ describe('Method get.accounts', () => {
 		});
 
 		it('returns delegate data by delegate name', async () => {
-			const { result } = await getAccounts({ username: accounts.delegate.username });
+			const { result } = await getAccounts({ username: accounts.delegate.delegate.username });
```
MichalTuleja, line 130: Why is that delegate object written twice? Did the API change?

nagdahimanshu: There was a bug in test. Its fixed now

MichalTuleja (approved):

PR #155, opened 2020-10-26
Migrate peer cache to blockchain data store - Closes #135

### LINKED ISSUE #135: Migrate peer cache to blockchain data store
### Description

Move peer search logic from controller to the blockchain data store. Keep the original structure of the way the data is retrieved. All peers need to be stored in-memory. Distinguish state and stateName.

### Motivation

- Storing data in the local database has a positive impact on performance
- It makes it easier to index and search data
- Proper code organisation makes it possible to use blockchain abstraction layer independently from Lisk Core version underneath

### Acceptance Criteria

- Peer data is available trough the same logic it is done before
- No need for a persistent PouchDB migration
- state and stateName are implemented internally
- The search logic is moved from the controller to the blockchain data store

sameersubudhi:
### What was the problem?
This PR resolves #135 

### How was it solved?
- [x] Peer data is available through the same logic it is done before
- [x] The search logic is moved from the controller to the blockchain datastore
- [x] state and stateName are implemented internally
- [x] Data need not be migrated to PouchDB for persistence

### How was it tested?
Local + Jenkins

### DIFF
--- a/services/core/methods/controllers/peers.js
+++ b/services/core/methods/controllers/peers.js
@@ -42,14 +42,12 @@ const getDisconnectedPeers = async params => {
 	};
 };
 
-const getPeersStatistics = () => {
-	const response = coreService.getPeersStatistics();
-
-	const meta = {};
+const getPeersStatistics = async () => {
+	const response = await coreService.getPeersStatistics();
 
 	return {
-		data: response,
-		meta,
+		data: response.data,
+		meta: response.meta,
 	};
 };
 
--- a/services/core/methods/controllers/peers.js
+++ b/services/core/methods/controllers/peers.js
@@ -34,7 +34,7 @@ const getConnectedPeers = async params => {
 };
 
 const getDisconnectedPeers = async params => {
-	const response = await coreService.getPeers(params);
+	const response = await coreService.getDisconnectedPeers(params);
 
 	return {
 		data: response.data,
--- a/services/core/methods/controllers/peers.js
+++ b/services/core/methods/controllers/peers.js
@@ -13,84 +13,19 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-const GeoService = require('../../shared/geolocation.js');
-const peerCache = require('../../shared/core/peerCache.js');
+const coreService = require('../../shared/core');
 
-const addLocation = async (ipaddress) => {
-	try {
-		const result = await GeoService.requestData(ipaddress);
-		return result;
-	} catch (e) {
-		return {};
-	}
-};
-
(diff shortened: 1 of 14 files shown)


nagdahimanshu (approved):

services/core/methods/controllers/peers.js
```diff
@@ -99,18 +34,22 @@ const getConnectedPeers = async params => {
 };
 
 const getDisconnectedPeers = async params => {
-	const response = await getPeers(Object.assign(params, { state: 'disconnected' }));
+	const response = await coreService.getPeers(params);
```
MichalTuleja, line 37: Should there be any constraint?

services/core/shared/core/compat/sdk_v3/mappings.js
```diff
@@ -15,6 +15,13 @@
 //  */
 // const { getDelegateRankByUsername } = require('./delegateCache');
 
+// const reverseMap = (originalMap) => {
+//     const result = {};
+//     Object.entries(originalMap).forEach(([k, v]) => result[v] = String(k).toLowerCase());
+
+//     return result;
+// };
+
```
MichalTuleja, line 24: Do we still need those lines?

services/gateway/sources/mappings/peer.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
 	os: '=,string',
 	version: '=,string',
 	state: '=,number',
+	stateName: '=,string',
```
MichalTuleja, line 23: @sameersubudhi @nagdahimanshu? `stateName` or `stateText`? What do you think?

MichalTuleja (changes requested):

services/core/methods/controllers/peers.js
sameersubudhi: Fixed the typo!

services/core/shared/core/compat/sdk_v3/mappings.js
sameersubudhi: Yes!

services/gateway/sources/mappings/peer.js
sameersubudhi: I prefer `stateName` for now.
If it weren't for the backward compatibility, I would prefer something like `connectionState` (text) and `connectionStateCode` (number).

MichalTuleja: Right, I'll keep this in mind for the next API design.

MichalTuleja (approved):

PR #156, opened 2020-10-27
Migrate delegate cache to data store - Closes #134

### LINKED ISSUE #134: Migrate delegate cache to data store
### Description

Migrate delegate cache to the data store and enable indexing and searching from the local database. Move any delegate-related logic to the blockchain abstraction layer and make use of the database. Keep delegate data in-memory. Migrate controllers to the blockchain data store.

### Motivation

- Storing data in the local database has a positive impact on performance
- It makes it easier to index and search data
- Proper code organisation makes it possible to use blockchain abstraction layer independently from Lisk Core version underneath

### Acceptance Criteria

- Delegate cache is migrated to PouchDB
- Search among delegates is possible by the PouchDB engine
- Delegate data are refreshed every 1 minute

sameersubudhi:
### What was the problem?
This PR resolves #134 

### How was it solved?
- [x] Delegate cache is migrated to PouchDB (in memory)
- [x] Search among delegates is possible by the PouchDB engine
- [x] Delegate data are refreshed every 1 minute

### How was it tested?
Local + Jenkins


services/core/shared/pouchdb.js
```diff
@@ -28,7 +29,13 @@ const connectionPool = {};
...
 const createDb = async (name, idxList = []) => {
 	logger.debug(`Creating/opening database ${name}...`);
-	const db = new PouchDB(name, { auto_compaction: true });
+
+	let db;
+	if (name.split('/')[1] === config.db.collections.delegates.name) {
+		db = new PouchDB(name, { adapter: 'memory', auto_compaction: true });
+	} else {
+		db = new PouchDB(name, { auto_compaction: true });
+	}
```
MichalTuleja, line 38: Please make the adapter configurable in the config and keep PouchDB independent from the product.

MichalTuleja (changes requested):

services/core/shared/core/delegates.js
```diff
@@ -0,0 +1,126 @@
...
+		&& (!params.username || delegate.username === params.username)
+		&& (!params.address || delegate.account.address === params.address)
+		&& (!params.publickey || delegate.account.publicKey === params.publickey)
+		&& (!params.secpubkey || delegate.account.secondPublicKey === params.secpubkey)
+	));
+	return relevantDelegates.length;
+};
+
+module.exports = {
+	reloadDelegateCache: reload,
```
nagdahimanshu, line 123: Instead of calling reload function, can we directly call `loadAllDelegates()`?

module.exports = {
           reloadDelegateCache: loadAllDelegates,

services/core/jobs/delegates.js
```diff
@@ -22,7 +22,7 @@ module.exports = [
 	{
 		name: 'refresh.delegates',
 		description: 'Keep the delegate list up-to-date',
-		schedule: '*/5 * * * *', // Every 5 min
+		schedule: '* * * * *', // Every 1 min
```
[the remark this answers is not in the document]
sameersubudhi: On mainnet, with 1864 delegates, the initial caching takes less than 8 secs on an average. The reload op takes less than 5 secs every time.
On testnet, with 724 delegates, the numbers are much lower at 3 secs (initial loading) and 1.5 secs (reload).

Scheduling the job every minute seems appropriate to me.

nagdahimanshu (approved):
Except for one comment, looks good to me!

services/core/shared/core/delegates.js
sameersubudhi: Even though the `reload` method simply redirects to the `loadAllDelegates` method, for now, I prefer declaring the `reload` method explicitly instead of only mapping the `loadAllDelegates` method in the `module.exports` to keep it clean and more intuitive for anyone who refers the code.

MichalTuleja (approved):

PR #194, opened 2020-11-16
Fix RPC calls registration issue - Closes #193

### LINKED ISSUE #193: Fix RPC calls registration issue
### Actual behavior
Currently, `get.voters` and `get.votes` does not work properly. Register RPC in a proper way to avoid backward compatibility issue.

### Expected behavior
`get.voters` and `get.votes` works properly

### Steps to reproduce
call `get.voters` and `get.votes`

nagdahimanshu:
### What was the problem?
This PR resolves #193 

### How was it solved?
- [x] Registered RPC calls in a proper way
- [x] RPC call get.votes  works properly.
- [x] RPC call get.voters works properly.
- [x] All versions works

### How was it tested?
Local + Jenkins


services/gateway/namespaces.js
```diff
@@ -33,9 +33,9 @@ const filterApis = (requiredApis, availableApis) => {
 };
 
 module.exports = filterApis(config.api.ws, {
-	'/rpc': () => registerApi('http-version1', { ...defaultConfig }),
-	'/rpc-v1': () => registerApi('http-version1', { ...defaultConfig }),
-	'/rpc-test': () => registerApi('http-test', { ...defaultConfig }),
+	'/rpc': () => registerApi(['http-version1', 'http-version1-compat'], { ...defaultConfig }),
+	'/rpc-v1': () => registerApi(['http-version1', 'http-version1-compat'], { ...defaultConfig }),
+	'/rpc-test': () => registerApi(['http-test'], { ...defaultConfig }),
```
MichalTuleja, line 38: I think this array does not have to be obligatory there, can you make it support also a plain string when a single name is there?

MichalTuleja (changes requested):

services/gateway/namespaces.js
nagdahimanshu: Done!

services/gateway/shared/registerRpcApi.js
```diff
@@ -26,15 +26,22 @@ const { validate } = require('./paramValidator');
...
-
-	const allMethods = Utils.requireAllJs(path.resolve(__dirname, `../apis/${apiName}/methods`));
+	if (typeof apiNames === 'string') apiNames = [apiNames];
+	apiNames.map(apiName => {
+		Object.assign(
+			allMethods,
+			Utils.requireAllJs(path.resolve(__dirname, `../apis/${apiName}/methods`)),
+		);
+		return apiName;
+	});
```
sameersubudhi, line 44: Perhaps we could simplify it to:

```suggestion
	apiNames.forEach(apiName => Object.assign(
		allMethods,
		Utils.requireAllJs(path.resolve(__dirname, `../apis/${apiName}/methods`)),
	));
```

sameersubudhi (approved):
LGTM 👍

services/gateway/shared/registerRpcApi.js
nagdahimanshu: Done!

MichalTuleja (approved):
Looks perfect, thanks!

PR #202, opened 2020-11-23
Drop default anyId behaviour - Closes #187

### LINKED ISSUE #187: Drop default anyId behaviour
### Description

Current version of Lisk Service allows to use anyId when looking for a certain account. Although it is very convenient to have, the API users barely can benefit it and it brings confusion when. Remove `anyId` support and replace it to accountId by default.

### Motivation

- `anyId` brings confusion to the way accounts are accessed. Compare lisk address `1L` and delegate registered as `1L`.

### Acceptance Criteria

- `anyId` behaviour is not used for address resolutions anymore

nagdahimanshu:
### What was the problem?
This PR resolves #187 

### How was it solved?
- [x]  anyId behaviour is drop
- [x]  Update `getAddressbyAny` to support `sender=<address>`, `sender=publickey:<publickey>` and `sender=username:<username>`
- [x]  Update blocks to retrieve using username
- [x]  Added tests

### How was it tested?
Local + Jenkins


services/core/methods/controllers/blocks.js
```diff
@@ -26,11 +26,18 @@ const getBlocks = async params => {
 	if (typeof params.height === 'number') {
 		params.height = `${params.height}`;
 	}
+	if (params.username) {
+		params.address = await Core.getAddressByUsername(params.username);
+		if (!params.address) return { status: NOT_FOUND, data: { error: `Account ID ${params.username} not found.` } };
```
sameersubudhi, line 31:
```suggestion
		if (!params.address) return { status: NOT_FOUND, data: { error: `Account ID corresponding to username: '${params.username}' not found.` } };
```

services/core/shared/core/compat/sdk_v2/index.js
```diff
@@ -135,12 +135,11 @@ const getAddressByAny = async param => {
 		const parsedAddress = parseAddress(param);
 		if (validateAddress(parsedAddress)
 			&& await confirmAddress(parsedAddress)) return parsedAddress;
-		if (validatePublicKey(param)) return getAddressByPublicKey(param);
-		return getAddressByUsername(param);
 	}
 
 	const [prefix, body] = separateParam(param);
-	return paramNames[prefix](body);
+	if (prefix || body) return paramNames[prefix](body);
```
sameersubudhi, line 141: Checking for `body` within the condition, doesn't make sense if `prefix` is returned to be `undefined`.

```suggestion
	if (prefix && body) return paramNames[prefix](body);
```
or:
```suggestion
	if (prefix) return paramNames[prefix](body);
```

sameersubudhi (changes requested):
Kindly consider the suggested changes.

services/core/methods/controllers/blocks.js
nagdahimanshu: Done!

services/core/shared/core/compat/sdk_v2/index.js
nagdahimanshu: Done!

sameersubudhi (approved):
Looks good 👍

MichalTuleja (changes requested):
The username param for blocks does not work.
`:9901/api/v1/blocks?username=cc001`

nagdahimanshu (commented):
Fixed `:9901/api/v1/blocks?username=cc001`.

services/core/shared/core/delegates.js
```diff
@@ -124,7 +124,8 @@ const loadAllDelegates = async () => {
 	await BluebirdPromise.map(
 		delegateList,
 		async delegate => {
-			await cacheRedisDelegates.set(delegate.address, delegate);
+			const delegateAddress = delegate.address || delegate.account.address;
```
MichalTuleja, line 127: Does it depend on lisk-sdk version? Maybe it is worth moving it to the respective compat directories?

nagdahimanshu: Yes, it is version dependent So, I moved that logic to `compat` layer. Thanks for the suggestion!

MichalTuleja (approved):

PR #208, opened 2020-12-01
Create abstraction layer for blockchain index - Closes #207

### LINKED ISSUE #207: Create abstraction layer for blockchain index
### Description

Create abstraction layer for blockchain index based on the experimental PouchDB implementation.

Consider the following backends that might be used in the future:

- Redis
- RediSearch
- PostgreSQL
- SQlite
- MongoDB

### Motivation

- Need to know the technical limitations of storing blockchain data index

### Acceptance Criteria

- Comments in the issue with results
- Draft implementation for Redis

nagdahimanshu:
### What was the problem?
This PR resolves #207 

### How was it solved?

- [x] Abstraction layer for database
- [x] Draft implementation for Redis on accounts
- [x] Implement indexing and secondary indexes in redis
- [x] limit, sort and offset in redis

### How was it tested?
Locally + Jenkinks


services/core/config.js
```diff
@@ -201,5 +203,5 @@ config.log.gelf = process.env.SERVICE_LOG_GELF || 'false';
 config.log.file = process.env.SERVICE_LOG_FILE || 'false';
 config.log.docker_host = process.env.DOCKER_HOST || 'local';
 config.debug = process.env.SERVICE_LOG_LEVEL === 'debug';
-
+config.defaultDB = 'redisdb';
```
MichalTuleja, line 206: What do you think about calling it just `redis`?
I think it corresponds the common usage and the official name is also `Redis`.

nagdahimanshu: I would totally agree with you. I will update it to `Redis` now.

nagdahimanshu: done!

PR #276, opened 2020-12-17
Fixed transaction date conversion - Closes #274

### LINKED ISSUE #274: Fix transaction date conversion
### Actual behavior

Timestamp returned from `/api/v1/transactions` endpoint is incorrect.

### Expected behavior

Timestamp returned from `/api/v1/transactions` endpoint is in a proper unix time format.

### Steps to reproduce

Retrieve `/api/v1/transactions`.

`v0.2.0-beta.1`

### Additional information

Possible solution might a removal of the line

```
transaction.unixTimestamp = await coreApi.getUnixTime(transaction.timestamp)
```

from the following code:

```
			if (!transaction.timestamp) {
				const txBlock = (await getBlocks({ height: transaction.height })).data[0];
				transaction.timestamp = txBlock.unixTimestamp;
			}
			transaction.unixTimestamp = await coreApi.getUnixTime(transaction.timestamp);
			return transaction;
		},
```

nagdahimanshu:
### What was the problem?
This PR resolves #274 

### How was it solved?

- [x] Refactored transaction to parse timestamp properly for all version
- [x] Move timestamp conversion logic to compat layer

### How was it tested?
Local

### DIFF
--- a/services/core/shared/core/compat/sdk_v4/transactions.js
+++ b/services/core/shared/core/compat/sdk_v4/transactions.js
@@ -18,6 +18,7 @@ const coreApi = require('./coreApi');
 const {
 	getBlockchainTime,
 	validateTimestamp,
+	getUnixTime,
 } = require('../common');
 
 const getTransactionsByBlock = async block => {
@@ -60,6 +61,16 @@ const getTransactions = async params => {
 		if (response.data) transactions.data = response.data;
 		if (response.meta) transactions.meta = response.meta;
 	}
+
+	transactions.data = await BluebirdPromise.map(
+		transactions.data,
+		async transaction => {
+			const txBlock = (await coreApi.getBlocks({ height: transaction.height })).data[0];
+			transaction.unixTimestamp = await getUnixTime(txBlock.timestamp);
+			return transaction;
+		},
+		{ concurrency: transactions.data.length },
+	);
 
 	transactions.meta.total = transactions.meta.count;
 	transactions.meta.count = transactions.data.length;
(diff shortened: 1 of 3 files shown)


sameersubudhi (approved):

services/core/shared/core/compat/sdk_v4/transactions.js
```diff
@@ -61,6 +62,19 @@ const getTransactions = async params => {
 		if (response.meta) transactions.meta = response.meta;
 	}
 
+	transactions.data = await BluebirdPromise.map(
+		transactions.data,
+		async transaction => {
+			if (!transaction.timestamp) {
+				const txBlock = (await coreApi.getBlocks({ height: transaction.height })).data[0];
+				transaction.timestamp = txBlock.timestamp;
+			}
```
MichalTuleja, line 71: Is there any case in SDKv4 that the transaction time is there already?

nagdahimanshu: In case of (params.fromTimestamp ||  params.totimeStamp), we already resolve timestamp when we call function `getTransactionsByBlock`. So, just to avoid an unnecessary call i added this check.

```
	if (params.fromTimestamp || params.toTimestamp) {
		const blocks = await coreApi.getBlocks(params);
		transactions.meta.count = 0;
		await BluebirdPromise.map(
			blocks.data,
			async (block) => {
				const blkTransactions = await getTransactionsByBlock(block);
				transactions.data = transactions.data.concat(blkTransactions.data);
				transactions.meta.count += blkTransactions.meta.count;
			},
			{ concurrency: blocks.data.length },
		);
	}
```

MichalTuleja (approved):

PR #290, opened 2021-01-04
Vote endpoints without a param causes wrong HTTP response - Closes #233

### LINKED ISSUE #233: Vote endpoints without a param causes wrong HTTP response
### Actual behavior

The `/votes_sent` and `/votes_received` endpoints return error HTTP 500 instead of 4xx when there are no params defined.

### Expected behavior

The `/votes_sent` and `/votes_received` endpoints return error HTTP 400 when the when there are no params defined.

### Steps to reproduce

Call `/votes_sent` without a param.

`0.2.0-beta.1`

sameersubudhi:
### What was the problem?
This PR resolves #233 

### How was it solved?
- [x] Add validation for required param from a pool of optional request parameters
- [x] Add validation to check if complementary parameters exist for a user-specified request param
- [x] Add corresponding integration test cases

### How was it tested?
Local + Jenkins

### DIFF
--- a/services/gateway/apis/http-version1/methods/votesSent.js
+++ b/services/gateway/apis/http-version1/methods/votesSent.js
@@ -30,8 +30,14 @@ module.exports = {
 		secpubkey: { optional: true, type: 'string', min: 64, max: 64 },
 		limit: { optional: true, min: 1, max: 100, type: 'number' },
 		offset: { optional: true, min: 0, type: 'number' },
-		onefrom: ['address', 'username', 'publickey', 'secpubkey'],
 	},
+	paramsRequired: true,
+	validParamPairings: [
+		['address'],
+		['username'],
+		['publickey'],
+		['secpubkey'],
+	],
 	get schema() {
 		const votesSchema = {};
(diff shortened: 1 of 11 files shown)


services/gateway/apis/http-version1/methods/votesSent.js
```diff
@@ -30,6 +30,7 @@ module.exports = {
 		secpubkey: { optional: true, type: 'string', min: 64, max: 64 },
 		limit: { optional: true, min: 1, max: 100, type: 'number' },
 		offset: { optional: true, min: 0, type: 'number' },
+		onefrom: ['address', 'username', 'publickey', 'secpubkey'],
```
MichalTuleja, line 33: This can be defined separately at the upper level property. Ideally as array of arrays that define one of the properties that must be set.

sameersubudhi: Done!

nagdahimanshu (approved):
Looks good!

MichalTuleja (approved):

PR #301, opened 2021-01-12
Retrieve accounts by address - Closes #239

### LINKED ISSUE #239: Retrieve accounts by address
### Description

Make Lisk Service possible to retireve accounts by ID. 

Add `get.accounts` API route to the Gateway.

Endpoint: `/api/v2/accounts`

### Motivation

- The basic retrieval by ID is essential for the further development

### Acceptance Criteria

- Accounts can be retrieved by ID
- Gateway endpoint is created

sameersubudhi:
### What was the problem?
This PR resolves #239 

### How was it solved?
- [x] Add endpoint: `/api/v2/accounts`
- [x] Add RPC support for API route `get.accounts`
- [x] Enable account retrieval by `address`
- [x] The response structure is set as expected

### How was it tested?
Local


services/core/shared/core/accountUtils.js
```diff
@@ -0,0 +1,160 @@
...
+const isStringType = value => typeof value === 'string';
+
+const parseAddress = address => isStringType(address) ? address.toUpperCase() : '';
+
+const validateAddress = address => isStringType(address) && address.match(/^[0-9]{1,20}[L|l]$/g);
+
+const validatePublicKey = publicKey => isStringType(publicKey) && publicKey.match(/^([A-Fa-f0-9]{2}){32}$/g);
+
+// Lisk Core API functions
+const confirmAddress = async address => {
```
nagdahimanshu, line 41: Given function is already defined in `compat` layer, do we still need it here?

services/core/shared/core/accountUtils.js
```diff
@@ -0,0 +1,160 @@
...
+};
+
+const confirmUsername = async username => {
+    if (!username || typeof username !== 'string') return false;
+    const result = await getDelegates({ username });
+    if (!Array.isArray(result.data) || isEmptyArray(result.data)) return false;
+    return result.data[0].username === username;
+};
+
+const confirmPublicKey = async publicKey => {
```
nagdahimanshu, line 54: Same as above

nagdahimanshu (approved):
Looks good apart from one comment!

services/core/shared/core/accountUtils.js
sameersubudhi: The idea behind introducing `accountUtils` was to reduce duplication of code by moving as much common code possible to the common layer. This is still a work in progress and would be finished by the time we finish implementing the advanced features related to the `/accounts` endpoint. Gradually, we would also like to remove many of these methods as they might not be necessary.

PR #309, opened 2021-01-19
Fix forgers list - Closes #306

### LINKED ISSUE #306: Fix forgers list
### Actual behavior

The next forger list is not sorted in a proper order, which displays it wrong in the client.

### Expected behavior

The next forger list is sorted in a proper order.

### Steps to reproduce

- Retrieve the next forgers list

MichalTuleja:
### What was the problem?

This PR resolves #306

### How was it solved?

- [x] Added event `forgers.change` 
- [x] Added mapping in the gateway for `update.forgers`

### How was it tested?

Local client

### DIFF
--- a/services/core/shared/core/delegates.js
+++ b/services/core/shared/core/delegates.js
@@ -192,8 +192,7 @@ const getNextForgers = async params => {
 	};
 
 	const offset = params.offset || 0;
-	let limit = params.limit || 103;
-	if (sdkVersion < 4) limit = 101;
+	const limit = params.limit || 10;
 
 	forgers.data = nextForgers.slice(offset, offset + limit);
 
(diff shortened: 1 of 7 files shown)


services/core/shared/core/delegates.js
```diff
@@ -192,7 +192,8 @@ const getNextForgers = async params => {
 	};
 
 	const offset = params.offset || 0;
-	const limit = params.limit || 10;
+	let limit = params.limit || 103;
+	if (sdkVersion < 4) limit = 101;
```
sameersubudhi, line 196: If limit specified in params, consider it else the default value.
```suggestion
	if (sdkVersion < 4) limit = params.limit || 101;
```

sameersubudhi (changes requested):
Please find my remark below.

services/core/shared/core/delegates.js
MichalTuleja: Rolled back to 10.

nagdahimanshu (approved):

sameersubudhi (approved):

PR #317, opened 2021-01-21
Fix non-indexed transaction calls - Closes # 316

### LINKED ISSUE #316: Fix non-indexed transaction calls
### Actual behavior

Transactions requests return wrong data when called with mixed non-indexed and indexed params.

### Expected behavior

Make the `transactions` endpoint able to return data on mixed params calls.

### Steps to reproduce

Request `/api/v1/transactions` with the following params: 
```
{
   type: 'REGISTERDELEGATE',
   sort: 'timestamp:desc',
}
```

sameersubudhi:
### What was the problem?
This PR resolves #316 

### How was it solved?
- [x] Translate transaction type param prior to index search
- [x] Assign default `fromTimestamp` and `toTimestamp` values, if missing, prior index search
- [x] Add indexing for delegate registrations
- [x] Filter out non-existent usernames in the response (`/voters`)

### How was it tested?
Local

### DIFF
--- a/services/core/config.js
+++ b/services/core/config.js
@@ -94,7 +94,7 @@ config.cacheNumOfBlocks = Number(process.env.CACHE_N_BLOCKS) || 202;
  * The block index may trigger indexing of other entities that are part of the block
  * such as transactions, accounts, votes etc.
  */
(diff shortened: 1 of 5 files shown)


nagdahimanshu (approved):
Looks good!

MichalTuleja (approved):

PR #328, opened 2021-01-27
Search index abstraction layer - Closes #296

### LINKED ISSUE #296: Define database access abstraction layer
### Description

The database abstraction layer is not clearly defined and all DB backends need to accept compatible parameters. Operation such as `write()`, `update()` and `getCount()` need to have clear definitions what they do.

### Motivation

This is needed to avoid compatibility issues between various database backends.

### Acceptance Criteria

- `upsert()`  is clearly defined and implemented
- `find()`  is clearly defined and implemented
- `delete()`  is clearly defined and implemented
- `getCount()`  is clearly defined and implemented

MichalTuleja:
### What was the problem?

This PR resolves #296

### How was it solved?

- [x] Created multi-purpose schema
- [x] Add MySQL backend
- [x] Migrate SDKv4 blocks to MySQL index
- [x] Migrate SDKv4 transactions to MySQL index
- [x] Trigger block update on each new block
- [x] Trigger transaction update on each new block
- [x] Implement MySQL index search by query
- [x] Implement MySQL query builder

### How was it tested?

- [x] Local
- [x] Jenkins


services/core/shared/indexdb/mysql.js
```diff
@@ -0,0 +1,188 @@
...
+
+	const upsert = async (inputRows) => {
+		let rawRows = inputRows;
+		if (!Array.isArray(rawRows)) rawRows = [inputRows];
+
+		const rows = [];
+		rawRows.forEach(item => {
+			const row = {};
+			Object.keys(schema).forEach(o => {
+				row[o] = cast(item[o], schema[o].type);
```
MichalTuleja, line 95: Add support for null

services/core/shared/indexdb/mysql.js
```diff
@@ -0,0 +1,188 @@
...
+
+		delete params.limit;
+		delete params.offset;
+
+		let res;
+		if (params.propBetween) {
+			const { propBetween } = params;
+			delete params.propBetween;
+			res = await knex.select(columns).table(tableName)
+				.whereBetween(propBetween.property, [propBetween.from, propBetween.to]);
```
MichalTuleja, line 137: add limit/offset

services/core/shared/core/compat/sdk_v4/schema/transactions.js
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+module.exports = {
+	primaryKey: 'id',
+	schema: {
+		id: { type: 'string' },
+		type: { type: 'integer' },
+		senderId: { type: 'string' },
+		recipientId: { type: 'string' },
+		amount: { type: 'integer' },
```
nagdahimanshu, line 24: value of `amount` is of type `BigInt`. (While writing amount to `mysql`, it throws an warning: `Error: ER_WARN_DATA_OUT_OF_RANGE: Out of range value for column 'amount' `)

Update to:
```
amount: { type: 'bigInteger' },
```

services/core/shared/indexdb/mysql.js
```diff
@@ -0,0 +1,183 @@
...
+	});
+
+	return knex;
+};
+
+const cast = (val, type) => {
+	if (type === 'number') return Number(val);
+	if (type === 'integer') return Number(val);
+	if (type === 'string') return String(val);
+	if (type === 'boolean') return Boolean(val);
```
nagdahimanshu, line 64: To cast `bigInteger`:

`if (type === 'bigInteger') return BigInt(val);`

services/core/shared/indexdb/mysql.js
```diff
@@ -0,0 +1,183 @@
...
+	if (await knex.schema.hasTable(tableName)) return knex;
+
+	await knex.schema
+	.createTable(tableName, table => {
+		Object.keys(schema).map(p => {
+			const kProp = (table[schema[p].type])(p);
+			if (schema[p].null === false) kProp.notNullable();
+			if (p === primaryKey) kProp.primary();
+			if (indexes[p]) kProp.index();
+			return kProp;
```
sameersubudhi, line 36: As a TODO, please consider adding support for composite primary keys and foreign keys in the future.

```suggestion
			const kProp = (table[schema[p].type])(p);
			if (schema[p].null === false) kProp.notNullable();
			if (p === primaryKey) kProp.primary();
			if (indexes[p]) kProp.index();
			
			// TODO: Add support for composite primary keys and foreign keys
			return kProp;
```

services/core/shared/indexdb/mysql.js
```diff
@@ -0,0 +1,183 @@
...
+	const find = (params, columns) => new Promise((resolve, reject) => {
+		queryBuilder(params, columns).then(response => {
+			resolve(response);
+		}).catch(err => {
+			logger.error(err.message);
+			reject(err);
+		});
+	});
+
+	const deleteIds = async ids => knex(tableName)
```
sameersubudhi, line 169: A more generic `delete` method could be put in place that builds atop the `find` method (or `queryBuilder`) and the resultSet chained by `del()`.

```suggestion
	const deleteIds = async ids => knex(tableName)
```

sameersubudhi (approved):
The current implemenation LGTM as is. 👍
However, kindly consider the following good-to-haves.

Also, I highly appreciate the idea and execution of `loadSchema` and `queryBuilder` in `services/core/shared/indexdb/mysql.js` 👏

services/core/shared/indexdb/mysql.js
MichalTuleja: The problem with delete is that `new` and `delete` are reserved keywords in JS and they confuse linter. I still doubt whether we really should use that. I kind of works, but still we need to avoid doing it I think.

https://www.w3schools.com/js/js_reserved.asp

sameersubudhi: Perhaps `findAndDelete`could be our way around?

MichalTuleja: I was thinking about it, although it is not needed yet.
We can implement on the first occurence.

sameersubudhi: Yes, sure. It was a recommendation for the future.

nagdahimanshu (approved):

PR #331, opened 2021-01-29
Fix block index build - Closes #330

### LINKED ISSUE #330: Fix block index build
### Actual behavior

All the existing blocks are not appropriately indexed when the application cold starts

### Expected behavior

All the blocks must be appropriately indexed when the application cold starts

### Steps to reproduce
Run Lisk Service against Lisk Core node version `3.0.0-beta.4` and verify the entries in the MySQL table `lisk.blocks`

SDKv5

sameersubudhi:
### What was the problem?
This PR resolves #330 

### How was it solved?
- [x] Fix block indexing logic at init
- [x] Avoid unnecessary multiple DB connection creation attempts at startup
- [x] Avoid overlaps while fetching blocks
- [x] Allow completion of DB migration before starting the index build
- [x] Add support for incomplete indexing upon restart

### How was it tested?
Local


services/core/shared/core/compat/sdk_v5/blocks.js
```diff
@@ -115,49 +115,66 @@ const buildIndex = async (from, to) => {
...
+			await buildIndex(blockIndexLowerRange, lowestIndexedHeight);
+		}
 	} catch (err) {
 		logger.warn('Unable to update block index');
 		logger.warn(err.message);
 	}
 };
 
-init();
+setTimeout(init, 5000);
```
MichalTuleja, line 177: We need to make sure this is replaced with the `init()` function, similarly to SDKv4.

MichalTuleja (approved):

PR #342, opened 2021-02-03
Fix transaction endpoint - Closes #340

### LINKED ISSUE #340: Fix transaction endpoint pagination and address param
### Actual behavior

The frontend team reposted two broken features:

- invalid pagination offset in `meta` 
- address param does not always work

### Expected behavior

- `meta` works correctly
- `address` param works correctly

### Steps to reproduce

`/api/v1/transactions?offset=10`
`/api/v1/transactions?address=`

nagdahimanshu:
### What was the problem?
This PR resolves #340 

### How was it solved?
- [x] meta works correctly
- [x] address param works correctly
- [x] implementation for counting records based on query params

### How was it tested?
Local

### DIFF
--- a/services/core/shared/core/compat/sdk_v4/transactions.js
+++ b/services/core/shared/core/compat/sdk_v4/transactions.js
@@ -34,12 +34,12 @@ const _getTrxFromCore = async params => {
 		transactions.data = await BluebirdPromise.map(
 			transactions.data,
 			async transaction => {
-				const response = await blockIdx.find(
+				const { resultSet } = await blockIdx.find(
 					{ id: transaction.blockId },
 					['timestamp', 'unixTimestamp']);
-				if (response.length > 0) {
-					transaction.timestamp = response[0].timestamp;
-					transaction.unixTimestamp = response[0].unixTimestamp;
+				if (resultSet.length > 0) {
+					transaction.timestamp = resultSet[0].timestamp;
+					transaction.unixTimestamp = resultSet[0].unixTimestamp;
 				}
 				return transaction;
 			},
@@ -94,8 +94,8 @@ const getTransactions = async params => {
 	}
 
 	// TODO: Add search by message
+	const { resultSet, total: [{ count }] } = await transactionIdx.find(params);
 
-	const resultSet = await transactionIdx.find(params);
 	if (resultSet.length > 0) {
 		const trxIds = resultSet.map(row => row.id);
 		transactions.data = await getTransactionByIds(trxIds);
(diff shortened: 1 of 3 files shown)


services/core/shared/indexdb/mysql.js
```diff
@@ -186,7 +186,7 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
 		.whereIn(primaryKey, ids)
 		.del();
 
-	const count = async () => knex(tableName).count({ count: 'id' });
+	const count = async () => knex(tableName).count({ count: '*' });
```
MichalTuleja, line 189: Why is this needed?

Btw, wouldn't it help to count by params?
The lines above return only all transactions in the db, without checking whether the index is finished.

nagdahimanshu: I realised now, its better to use '*'. Changing it back. Thanks

sameersubudhi: Using a single column based count query would be more efficient. Kindly consider reverting the change.
```suggestion
	const count = async () => knex(tableName).count({ count: 'id' });
```

services/core/shared/core/compat/sdk_v4/transactions.js
```diff
@@ -95,15 +96,15 @@ const getTransactions = async params => {
...
+	const [{ count }] = await transactionIdx.count();
 	if (resultSet.length > 0) {
 		const trxIds = resultSet.map(row => row.id);
 		transactions.data = await getTransactionByIds(trxIds);
 	}
 
-	transactions.meta.total = transactions.meta.count;
 	transactions.meta.count = transactions.data.length;
-	transactions.meta.offset = params.offset || 0;
+	transactions.meta.total = count;
```
sameersubudhi, line 105: Assigning the absolute `count` value as `total` might not be the correct approach. If someone queries `/api/v1/transactions?type=10`, then not all the transactions in the DB would be fitting this user request. Hence, we'd be presenting the user an intuition that a lot more transactions of type `10` exist than actual.

Please consider performing a `COUNT('id')` for the request parameters that the user sends and returning that value. We can discuss this further offline.

sameersubudhi (changes requested):
Please find my remarks below.

services/core/shared/indexdb/mysql.js
nagdahimanshu: Updated

services/core/shared/core/compat/sdk_v4/transactions.js
nagdahimanshu: Implemented the logic for it.

sameersubudhi (approved):
Looks good! 👍

services/core/shared/indexdb/mysql.js
```diff
@@ -141,36 +141,46 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
...
-			.limit(limit)
-			.offset(offset);
+		if (params.orWhere) {
+			const { orWhere } = params;
+			delete params.orWhere;
+			query.where(params).orWhere(orWhere);
+			countQuery.where(params).orWhere(orWhere);
+		}
+
+		const [{ count }] = await countQuery.andWhere(params);
```
MichalTuleja, line 175: That implementation looks nice, but I'd rather not change the original API.
Can you please move the count logic to the `count(params)` function?
Then we could pass the same params in the app logic.

MichalTuleja (changes requested):

PR #344, opened 2021-02-04
Test suite improvements - Closes #343

### LINKED ISSUE #343: Improve SDKv4 test capabilities
### Description

The SDKv4 implementation is missing integration tests.

Make sure the tests for the API version 1 are performing well on a local machine.

### Motivation

- The QA process needs to be improved

### Acceptance Criteria

- Tests are updated to the betanet env
- Tests are passing

MichalTuleja:
### What was the problem?
This PR resolves #343

### How was it solved?

- [x] Make directory for APIv1
- [x] Distinguish SDKv4 from SDKv2 in config files
- [x] Second public key tests
- [x] Delegates: Fix failing tests with standby delegates on betanet (n<100)
- [x] Delegates: Add tests for standby delegate filtering
- [x] Accounts: Rank bug
- [x] Accounts: Incoming/outgoing transactions
- [x] Received votes: Update tests
- [x] Sent votes: Update tests
- [x] Transactions: Fetch all transactions in genesis block
- [x] Linter fixes
- [x] Transactions query with min/max amount
- [x] Fix readiness reporting

### How was it tested?

Local + Jenkins


services/core/shared/indexdb/mysql.js
```diff
@@ -165,7 +165,9 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
 
 		if (params.orWhere) {
 			const { orWhere } = params;
-			query.where(queryParams).orWhere(orWhere);
+			query.where(function () {
```
MichalTuleja, line 168: Does it fail on linter? Wouldn't `() => {}` work as well?

nagdahimanshu: It is not possible to switch it to `arrow functions` in this case because Arrow functions doesn't have their own `this` binding and we need to use `this` for the results. Also, it doesn't fail on linter so we are good now.

services/core/methods/controllers/accounts.js
```diff
@@ -31,22 +32,27 @@ const getDataForAccounts = async params => {
 	response.meta = {};
 	response.links = {};
 
+	const accountDataCopy = JSON.parse(JSON.stringify(accounts.data));
```
sameersubudhi, line 35: Do we need to clone the response? 
This would fail if `accounts.data` contained any `BigInt` values. `JSON.stringify` doesn't handle them by default.

services/core/methods/controllers/accounts.js
```diff
@@ -31,22 +32,27 @@ const getDataForAccounts = async params => {
...
+					account);
+					account.knowledge = await getAccountKnowledge(account.address);
+			} catch (err) {
+				logger.warn(err.message);
+			}
+		}, { concurrency: 4 });
 
-		response.meta.count = response.data.length;
+		response.data = accountDataCopy;
+		response.meta.count = accounts.data.length;
```
sameersubudhi, line 55: For consistency purposes.
```suggestion
		response.meta.count = accountDataCopy.length;
```

services/core/shared/core/accounts.js
```diff
@@ -29,7 +29,9 @@ const getAccounts = async params => {
 	await Promise.all(accounts.data.map(async account => {
 		if (account.isDelegate === true) {
 			const delegate = await getDelegates({ address: account.address });
-			account.delegate = { ...account.delegate, ...delegate.data[0] };
+			const delegateOrigProps = JSON.parse(JSON.stringify(account.delegate));
+			const delegateExtraProps = JSON.parse(JSON.stringify(delegate.data[0]));
```
sameersubudhi, line 33: Same as above.

sameersubudhi (changes requested):
Please find my remarks below.

services/core/methods/controllers/accounts.js
MichalTuleja: It should map them to `string` shouldn't it?

MichalTuleja: Without it rank is always `NaN`, for some reason these properties are affected by passing a reference.

sameersubudhi: `JSON.stringify` doesn't automatically parse it to a string. We could utilize the `parseToJSONCompatObj` utility from `services/core/shared/core/compat/common/utils.js` in `development` branch.

nagdahimanshu (approved):

MichalTuleja:
Merging despite CI fails, the SDKv4 tests are passing locally.
SDKv2 CI will be fixed in #353

sameersubudhi (approved):

PR #362, opened 2021-02-16
Update transaction index on block arrival - Closes #248

### LINKED ISSUE #248: Update transaction index on block arrival
### Description

Each block that contains transactions needs to be retrieved from Lisk Core and indexed in the transaction database.

### Motivation

- Proper event propagation is required by UI clients

### Acceptance Criteria

- A new block with transactions triggers the event
- Subscribe API is updated

nagdahimanshu:
### What was the problem?
This PR resolves #248 

### How was it solved?

- [x] A new block with transactions triggers the event
- [x] Update transaction indexes on new block arrival
- [x] Refactor event to pass new block SDKv5 (Right now it pass only block id)
- [x] Event `transactions.confirmed` works properly if new transactions comes
- [x] Added `numberOfTransactions` in a response
- [x] Added `app:chain:validators:change` to support `round.change` event in service

### How was it tested?
Local


services/core/shared/core/compat/sdk_v5/transactions.js
```diff
@@ -208,7 +208,19 @@ const getTransactions = async params => {
...
+	const transactions = await getTransactions({ blockId });
+	return {
+		data: transactions,
+		meta: {
+			offset: 0,
+			count: transactions.length,
+		},
+	};
+};
+
```
sameersubudhi, line 221: `getTransactionsByBlockId` can be simply reduced to:
The standard response envelope (Line 174, `const transactions`) is already defined inside `getTransactions`.

```suggestion
const getTransactionsByBlockId = async blockId =>  getTransactions({ blockId });

```

sameersubudhi (changes requested):
Please find my feedback below.

services/core/shared/core/compat/sdk_v5/transactions.js
nagdahimanshu: Done!

services/core/shared/core/compat/sdk_v5/transactions.js
```diff
@@ -208,7 +208,10 @@ const getTransactions = async params => {
 	return transactions;
 };
 
+const getTransactionsByBlockId = blockId => getTransactions({ blockId });
```
sameersubudhi, line 211: `getTransactionsByBlockId` could be bettered by setting the `limit` param in the request as well, for scenarios where the blocks contain over 10 transactions (default limit value). Let us discuss this offline.

sameersubudhi (changes requested):
Let us discuss the following:

services/core/shared/core/compat/sdk_v5/transactions.js
nagdahimanshu: Refactor `getTransactionsByBlockId` to fetch new block and transactions directly from `lisk-core`.

sameersubudhi (approved):
LGTM! Thank you 😊

services/core/shared/core/compat/sdk_v5/transactions.js
```diff
@@ -208,7 +208,34 @@ const getTransactions = async params => {
...
 
+const getTransactionsByBlockId = async blockId => {
+	const [block] = (await coreApi.getBlockByID(blockId)).data;
+	const transactions = await BluebirdPromise.map(
+		block.payload,
+		async transaction => {
+			transaction.unixTimestamp = block.header.timestamp;
+			transaction.height = block.header.height;
+			transaction.blockId = block.header.id;
+			const account = await getIndexedAccountInfo({ publicKey: transaction.senderPublicKey });
```
MichalTuleja, line 219: Is there any reason to get account by public key instead of address?

nagdahimanshu: There is no specific reason here. The only reason is we are retrieving blocks directly from `lisk-core` and sender address is not available in `block.payload`.  We can resolve `address` from `public key` and then get `account` by address but anyway, we are resolving missing information in transaction for specific account(username & address).
```
 const account = await getIndexedAccountInfo({ publicKey: transaction.senderPublicKey });
 transaction.senderId = account && account.address ? account.address : undefined;
 transaction.username = account && account.username ? account.username : undefined;
```

MichalTuleja (approved):

Issue #372, opened 2021-02-24
Establish test blockchain for CI


MichalTuleja:
### Description

Current test suite does is missing blocks and transaction in the test blockchain.

Make a snapshot of Lisk Core blockchain and integrate it with out CI process.

### Motivation

- Stable and reliable test supported by automation are crucial for quality development process

### Acceptance Criteria

- Blockchain snapshot is pushed to the branch
- Blockchain snapshot is being used in CI

Issue #262, opened 2020-12-15
Search index dump and restore tool


MichalTuleja:
### Description

Create an independent script that makes possible to dump and pack the blockchain search index.
Create a tool that restores that database from the dumped files.

### Motivation

- Providing dump and restore tool makes it possible to speed-up the coldstart process.

### Acceptance Criteria

- Dump script is ready
- Restore script is ready
- Versioning support

PR #373, opened 2021-02-24
Add extended attribute calculation for delegate - Closes #261

### LINKED ISSUE #261: Add extended attribute calculation for delegate accounts
### Description

Calculate the following data from the accounts and transactions.

- producedBlocks
- rank
- rewards

In case any of them requires index starting from height 1, make sure it is calculated in a job queue.
Those properties should be enabled only if Lisk Service configuration allows caching to the height = 1.

Endpoint: `/api/v2/accounts`

### Motivation

- This feature is required by the UI clients

### Acceptance Criteria

- Extended attributes are calculated
- Retrieval of extended attributes is possible
- Gateway is updated with the API version 2

#### Additional information

- `isDelegate` attribute has to be `true`
- The results have to be calculated in Lisk Service
- The rank has to be based on voting power, including the allowed amount of self-votes

nagdahimanshu:
### What was the problem?
This PR resolves #261

### How was it solved?

- [x] Map status and rank in delegate response
- [x] Support status and query delegates by status
- [x] Support status and query accounts by status
- [x] Support search param and query delegates by search param (wildcard search)
- [x] Enable integration tests for following scenarios
  - [x] Delegate filter by status
  - [x] Delegate filter by search param
- [x] Extended attributes are calculated
  - [x] rank
  - [x] producedBlocks
  - [x] rewards
- [x] Retrieval of extended attributes is possible
  - [x] rank based sorting for delegates
- [x] Gateway is updated with the API version 2
- [x] Properties (`producedBlocks` and `rewards`) only enabled when indexing full blockchain

### How was it tested?
Local + Jenkins


services/core/shared/core/compat/sdk_v5/accounts.js
```diff
@@ -139,6 +139,16 @@ const getAccountsFromCore = async (params) => {
 
 const getAccounts = async params => {
 	const accountsDB = await getAccountsIndex();
+	if (params.sort && params.sort.includes('rank')) {
+		return new Error('Rank based sorting is only supported along delegates accounts');
+	}
```
sameersubudhi, line 144: Instead of returning an error, perhaps we should consider redirecting the call to `getDelegates`, to set the user expectations consistent.

services/core/shared/indexdb/mysql.js
```diff
@@ -196,6 +196,10 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
 			query.where(`${property}`, 'like', `%${pattern}%`);
 		}
 
+		if (params.aggregate) {
+			query.where(queryParams).sum(params.aggregate);
```
sameersubudhi, line 200:
```suggestion
			query.where(queryParams).sum(`${params.aggregate} as total`);
```

services/core/shared/core/compat/sdk_v5/delegates.js
```diff
@@ -14,25 +14,38 @@
...
         delegate.account = {
             address: delegate.address,
             publicKey: delegate.publicKey,
         };
-
+        if (getIsSyncFullBlockchain()) {
+            const [{ rewards }] = await blocksDB.find({
+                generatorPublicKey: delegate.publicKey, aggregate: 'reward as rewards',
+            });
+            delegate.rewards = rewards;
```
sameersubudhi, line 44:
```suggestion
            const [{ total }] = await blocksDB.find({
                generatorPublicKey: delegate.publicKey, aggregate: 'reward',
            });
            delegate.rewards = total;
```

services/core/shared/core/compat/sdk_v5/delegates.js
```diff
@@ -14,25 +14,38 @@
...
         meta: {},
     };
-
     const punishmentHeight = 780000;
     const response = await getAccounts({ isDelegate: true, limit: params.limit });
     if (response.data) delegates.data = response.data;
     if (response.meta) delegates.meta = response.meta;
 
-    delegates.data.map(delegate => {
+    delegates.data.map(async delegate => {
```
sameersubudhi, line 34: Please consider using BluebirdPromise.

services/gateway/sources/version2/mappings/account.js
```diff
@@ -63,10 +63,11 @@ module.exports = {
 			isBanned: '=,boolean',
 			totalVotesReceived: '=,string',
 			missedBlocks: '=,number',
-			producedBlocks: '=,number',
+			producedBlocks: 'producedBlocks,number',
```
sameersubudhi, line 66: Why do we need this change? Both the property names are the same. 
```suggestion
			producedBlocks: '=,number',
```

services/gateway/sources/version2/mappings/account.js
```diff
@@ -63,10 +63,11 @@ module.exports = {
...
 			productivity: '=,string',
-			rank: '=,number',
-			rewards: '=,string',
+			rank: 'rank,number',
+			status: 'status,string',
+			rewards: 'rewards,string',
```
sameersubudhi, line 70: Same as above!

tests/integration/api_v2/http/accounts.test.js
```diff
@@ -196,4 +196,26 @@ describe('Accounts API', () => {
 			expect(response.meta).toMap(metaSchema);
 		});
 	});
+
+	describe('Retrieve accounts based on status', () => {
+		it('returns delegate accounts when status is active', async () => {
```
sameersubudhi, line 201: Please consider adding test with multiple status values, e.g.: `status=active,standby` or similar values.

Jenkinsfile
```diff
@@ -108,6 +108,7 @@ pipeline {
 						set +e
 						while [ $ready -ne 0 ]; do
 							curl --fail --verbose http://127.0.0.1:9901/api/v2/blocks
+							curl --fail --verbose http://127.0.0.1:9901/api/v2/accounts
```
sameersubudhi, line 111: Why did we shift from `/api/v2/blocks` to `/api/v2/accounts`?

We should change to `/api/ready` instead. But this would only be doable after we finish: #262 or add dummy transactions after the container build phase during the CI.
```suggestion
							curl --fail --verbose http://127.0.0.1:9901/api/ready
```

sameersubudhi (changes requested):
Please find my review comments.

Jenkinsfile
MichalTuleja: I guess that waiting for accounts (or transactions) gives us more time to build an index during initial start. The `/api/ready` endpoint reports gateway readiness only.

sameersubudhi: Then, perhaps waiting for `/api/v2/votes` is the best check to have here as it would ensure that all the indexes have started building.

nagdahimanshu: I think for now `/api/v2/accounts` is enough as we dont have any transactions available in `Jenkins`. Once #372 is implemented, we can enable check either for `votes` or `transactions`.

services/gateway/sources/version2/mappings/account.js
nagdahimanshu: Clarified offline!

services/core/shared/core/compat/sdk_v5/accounts.js
sameersubudhi: Let us keep it as is unless explicitly requested.

sameersubudhi (approved):
LGTM!

MichalTuleja (approved):

PR #376, opened 2021-02-25
Make Jenkins CI run against Betanet 5

Nazgolze:
### What was the problem?
This PR changes the Jenkins job to run against betanet5 

### How was it solved?

Modified Jenkinsfile.deployment, added relevant betanet5 docker compose and env files

### How was it tested?

Tested on Jenkins

### Notes
Also adds workaround for https://github.com/docker/for-linux/issues/264

### DIFF
--- a/Jenkinsfile.deployment
+++ b/Jenkinsfile.deployment
@@ -39,13 +39,13 @@ pipeline {
 			when { branch 'development' }
 			environment {
 				ENABLE_HTTP_API='http-version1,http-version1-compat,http-status,http-test'
-				ENABLE_WS_API='rpc,rpc-v1,rpc-test,blockchain'
+				ENABLE_WS_API='rpc,rpc-v1,rpc-test,blockchain,rpc-v2'
 			}
 			steps {
 				ansiColor('xterm') {
 					dir('docker') {
-						sh 'make -f Makefile.deployment.betanet mrproper'
-						sh 'make -f Makefile.deployment.betanet up'
+						sh 'make -f Makefile.deployment.betanet5 mrproper'
+						sh 'make -f Makefile.deployment.betanet5 up'
 					}
 				}
 			}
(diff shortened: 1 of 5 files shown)


docker/lisk_service/docker-compose.core.yml
```diff
@@ -14,6 +14,8 @@ services:
     env_file:
       - ./lisk_service/env/common.env
       - ./lisk_service/env/core.env
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
```
fchavant, line 18: Where does `host-gateway` come from?

Nazgolze: It's more magic :grin:

Nazgolze: I think it's pretty new too. Might be the very latest version of Docker even

fchavant: And this one only works on linux? Or why do we not use it directly in `LISK_CORE_*`?

Jenkinsfile.deployment
```diff
@@ -39,17 +39,17 @@ pipeline {
 			when { branch 'development' }
 			environment {
 				ENABLE_HTTP_API='http-version1,http-version1-compat,http-status,http-test'
-				ENABLE_WS_API='rpc,rpc-v1,rpc-test,blockchain'
+				ENABLE_WS_API='rpc,rpc-v1,rpc-test,blockchain,rpc-v2'
```
MichalTuleja, line 42: `ENABLE_WS_API='http-status,http-test,blockchain,http-version2,rpc-v2'`

MichalTuleja (changes requested):

docker/lisk_service/docker-compose.core.yml
Nazgolze: This looks like the original PR https://github.com/moby/moby/pull/40007 
It's weird, but it looks like `host-gateway` can't just be used directly

fchavant (approved):

MichalTuleja (approved):

PR #378, opened 2021-02-25
Improve block indexing - Closes #360

### LINKED ISSUE #360: Improve block indexing
### Description

Improve SDKv5 block indexing.

- Migrate to events signals
- Use MySQL index to determine which blocks are missing

### Motivation

- Current implementation is missing proper event handling

### Acceptance Criteria

- Blocks are retrieved using MySQL
- Internal signals are passed
- Tests are written

sameersubudhi:
### What was the problem?
This PR resolves #360 

### How was it solved?
- [x] Extend `mysql.js` to support raw SQL queries
- [x] Add utility to determine missing blocks between a certain height range
- [x] Fix `buildIndex` method to allow indexing when `from` and `to` heights are equal (index single block)
- [x] Index the block finality status alongside the existing information
- [x] (Re-)index an incoming block only when it is not indexed or already existent but not finalised
- [x] Update indexed blocks' finality status upon new block arrival
- [x] Index handling upon fork detection for the below entities:
  - [x] `blocks`
  - [x] `transactions`
  - [x] `votes`
- [x] Add index ready flag
- [x] Add support for index ready status

### How was it tested?
Local

### DIFF
--- a/services/core/shared/indexdb/mysql.js
+++ b/services/core/shared/indexdb/mysql.js
@@ -253,9 +253,8 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
 		return totalCount.count;
 	};
 
-	const rawQuery = async (queryStatement) => {
-		const result = await knex.raw(queryStatement);
-		const [resultSet] = result;
+	const rawQuery = async queryStatement => {
+		const [resultSet] = await knex.raw(queryStatement);
 		return resultSet;
 	};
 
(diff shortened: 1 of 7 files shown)


services/core/shared/indexdb/mysql.js
```diff
@@ -253,11 +253,18 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
 		return totalCount.count;
 	};
 
+	const rawQuery = async (queryStatement) => {
+		const result = await knex.raw(queryStatement);
+		const [resultSet] = result;
+		return resultSet;
+	};
```
nagdahimanshu, line 260:
```suggestion
	const rawQuery = async (queryStatement) => {
		const [resultSet] = await knex.raw(queryStatement);
		return resultSet;
	};
```

nagdahimanshu (approved):
LGTM!

MichalTuleja (approved):

PR #382, opened 2021-03-03
Fix vote weight calculation - Closes #381

### LINKED ISSUE #381: Fix vote weight calculation
### Actual behavior

Vote weight calculation uses regular `Number` type.

### Expected behavior

Vote weight calculation requires using `BigInt`.

Make sure big integers are used where needed for aggregated data.

### Steps to reproduce

Retrieve vote weight for delegates with a great voting power.

sameersubudhi:
### What was the problem?
This PR resolves #381 

### How was it solved?
- [x] Delegate `voteWeight` calculation logic now uses `BigInt` instead of `Number` for both:
  - [x] SDKv4
  - [x] SDKv5
- [x] Parse `BigInt` values to `String` type before serialization
- [x] Fix `totalVotesReceived` for delegates in both:
  - [x] SDKv4
  - [x] SDKv5
- [x] Added `total` in meta for accounts response
- [x] Update logic for `update.block` event
- [x] Reload the list of all delegates with every block arrival for Rank and Status re-computation,

### How was it tested?
Local

### DIFF
--- a/services/core/events/blockchain.js
+++ b/services/core/events/blockchain.js
@@ -60,6 +60,7 @@ module.exports = [
 		description: 'Track round change updates',
 		controller: callback => {
 			signals.get('newBlock').add(async () => {
+				await core.reloadDelegateCache();
 				await core.reloadNextForgersCache();
 				const forgers = await core.getNextForgers({ limit: 25 });
 				callback(forgers);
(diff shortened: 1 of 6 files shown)


sameersubudhi (commented):
Thank you. LGTM! 👍

MichalTuleja (approved):

PR #385, opened 2021-03-04
Enable transaction broadcast - Closes #259

### LINKED ISSUE #259: Enable transaction broadcast
### Description

Make it possible to broadcast transactions using Lisk Service as a proxy.

Use `app:postTransaction` method to broadcast an encoded transaction.

Make sure so the client gets instant response from the peer.

### Motivation

- This feature is required by the UI clients

### Acceptance Criteria

- Transaction broadcast is possible
- Gateway is updated with the API version 2

#### Endpoints

- HTTP POST `/api/v2/transactions​`
- RPC `post.transactions`

#### Request parameters

No params required.

Payload:

```jsonc
{"transaction":"08021000180d2080c2d72f2a200fe9a3f1a21b5530f27f87a414b549e79a940bf24fdf2b2f05e7f22aeeecc86a32270880c2d72f12144fd8cc4e27a3489b57ed986efe3d327d3de40d921a0a73656e6420746f6b656e3a4069242925e0e377906364fe6c2eed67f419dfc1a757f73e848ff2f1ff97477f90263487d20aedf538edffe2ce5b3e7601a8528e5cd63845272e9d79c294a6590a"}
```

#### Response example

200 OK

```jsonc
{
  "message": "Transaction payload was successfully passed to the network node"
  "transactionId": "123456789"
}
```

400 Bad Request
```jsonc
{
  "error": true,
  "message": "Transaction payload was rejected by the network node"
}

```

500 Internal Server Error
```jsonc
{
  "error": true,
  "message": "Unable to reach a network node"
}

```

sameersubudhi:
### What was the problem?
This PR resolves #259 

### How was it solved?
- [x] Configure the Access-Control-Allow-Methods CORS header to only permit `GET` and `POST` requests
- [x] Add gateway support the following endpoints:
  - [x] HTTP: `POST /transactions`
  - [x] RPC: `post.transactions`
- [x] Validate the transaction binary payload to be in `hex` format
- [x] Add support in core to submit the signed transaction binary to the Lisk Core, sent within the request body
- [x] Delay delegate cache initialization at init to avoid indexing errors

### How was it tested?
Local

### DIFF
--- a/services/core/jobs/delegates.js
+++ b/services/core/jobs/delegates.js
@@ -14,9 +14,14 @@
  *
  */
 const logger = require('lisk-service-framework').Logger();
-
 const core = require('../shared/core');
-// const delegateCache = require('../shared/core/delegateCache');
+
+const setTimeoutPromise = ms => new Promise(resolve => setTimeout(resolve, ms));
+
+const sleepAndExec = async (fn, delay = 1000) => {
+	await setTimeoutPromise(delay);
+	return fn();
+};
 
 module.exports = [
 	{
@@ -26,7 +31,7 @@ module.exports = [
 		updateOnInit: true,
 		init: async () => {
 			logger.debug('Scheduling initial list update...');
-			await core.reloadDelegateCache();
+			await sleepAndExec(core.reloadDelegateCache);
 			await core.reloadNextForgersCache();
 		},
 		controller: () => {
--- a/services/gateway/apis/http-version2/methods/postTransactions.js
+++ b/services/gateway/apis/http-version2/methods/postTransactions.js
@@ -25,10 +25,6 @@ module.exports = {
 	params: {
 		transaction: { optional: false, type: 'string', min: 1 },
 	},
-	paramsRequired: true,
-	validParamPairings: [
-		['transaction'],
-	],
 	source: transactionsSource,
 	envelope,
 };
--- a/services/gateway/apis/http-version2/methods/postTransactions.js
+++ b/services/gateway/apis/http-version2/methods/postTransactions.js
@@ -22,7 +22,7 @@ module.exports = {
 	rpcMethod: 'post.transactions',
 	tags: ['Transactions'],
 	params: {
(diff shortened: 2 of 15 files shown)


services/core/jobs/delegates.js
```diff
@@ -26,7 +31,7 @@ module.exports = [
 		updateOnInit: true,
 		init: async () => {
 			logger.debug('Scheduling initial list update...');
-			await core.reloadDelegateCache();
+			await sleepAndExec(core.reloadDelegateCache);
```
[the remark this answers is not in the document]
sameersubudhi: The delay is necessary only for the initialization scenario when the tables are getting created in the database. Since the block indexing also begins simultaneously, and the 'blocks' table doesn't exist yet, both this and the `init` method in blocks try to create the table, which results in an error and causes the indexing process to exit prematurely.

nagdahimanshu (approved):
LGTM!

services/gateway/apis/http-version2/methods/postTransactions.js
```diff
@@ -0,0 +1,28 @@
...
+const transactionsSource = require('../../../sources/version2/postTransactions');
+
+module.exports = {
+	version: '2.0',
+	swaggerApiPath: '/transactions',
+	httpMethod: 'POST',
+	rpcMethod: 'post.transactions',
+	tags: ['Transactions'],
+	params: {
+		transaction: { optional: false, type: 'string', min: 1, pattern: /^\b[0-9a-fA-F]+\b$/ },
```
MichalTuleja, line 25: Does the body JSON also count as a param? Or is it the URL param?

sameersubudhi: Moleculer framework by default appends all the body JSON properties and request params into `$req.params`.
So, in our case `transaction` param is supposed to be part of the JSON object sent within the request body.

MichalTuleja (approved):

Issue #250, opened 2020-12-15
Retrieve accounts by search index


MichalTuleja:
### Description

Make the version 2 API able to retrieve data by various criteria.

Make sure that additional conditions are met according to the specification below.

Validation rules, intervals and type checking need to be performed at the Gateway level.

### Motivation

- Random access and search by those properties is required by the UI

### Acceptance Criteria

- Version 2 API supports additional search parameters

#### Endpoints

- HTTP `/api/v2/accounts`
- RPC `get.accounts`

#### Request parameters

Make the version 2 API able to retrieve data by those criteria.

Parameter | Type | Validation | Default | Comment
-- | -- | -- | -- | --
address | String | `/^lsk([a-hjkm-z]\|[2-9]){38}$//^[1-9]\d{0,19}[L\|l]$/` | *(empty)* | Resolves new and old address system
publicKey | String | `/^([A-Fa-f0-9]{2}){32}$/` | *(empty)* |  
username | String | `/^[a-z0-9!@$&_.]{1,20}$/` | *(empty)* |  
limit | Number | `<1;100>` | 10 |  
offset | Number | `<0;+Inf>` | 0 |  
sort | Array of strings | `[“balance:asc”, “balance:desc”, “rank:asc”, “rank:desc”]` | `balance:desc` | Rank is dedicated to delegate accounts

#### Response example

200 OK

```jsonc
{
  "data": {
    "summary": {
      "address": "02a1de92405edeaac13913bd089b07eb73cbd762",
      "balance": "151146419900",
      "username": "liberspirita",
      "publicKey": "968ba2fa993ea9dc27ed740da0daf49eddd740dbd7cb1cb4fc5db3a20baf341b",
      "isDelegate": true,
      "isMultisignature": true,
    },
    "knowledge": {
      "owner": "Genesis Account",
      "description": ""
    },
    "token": {
      "balance": "151146419900"
    },
    "sequence": {
      "nonce": "11"
    },
    "keys": {
      "numberOfSignatures": 0,
      "mandatoryKeys": [],
      "optionalKeys": [],
      "members": [
        {
          "address": "lsk24cd35u4jdq8szo3pnsqe5dsxwrnazyqqqg5eu",
          "publicKey": "968ba2fa993ea9dc27ed740da0daf49eddd740dbd7cb1cb4fc5db3a20baf341b",
          “isMandatory”: true,
        }
      ],
      "memberships": [
        {
[...]

PR #401, opened 2021-03-17
Transactions generated by a delegate in test blockchain - Closes #398

### LINKED ISSUE #398: Transactions generated by a delegate in test blockchain
### Description

Add transactions generated by a delegate in the test blockchain.

Make sure they are possible to find by `username`.

Enable and update tests that cover this scenario.

### Motivation

- It is important to test the API completely

### Acceptance Criteria

- Required transactions are added
- Tests are updated

sameersubudhi:
### What was the problem?
This PR resolves #398 

### How was it solved?
- [x] Create a new account on the blockchain
- [x] Transfer funds to the new account
- [x] Register the new account as a delegate
- [x] Send outgoing transactions from the newly registered delegate account
- [x] Add more accounts/transactions/delegate registrations to the blockchain via the stress test script
- [x] Refactor the transactions integration tests
- [x] Enable the transactions integration tests
- [x] Wait until the jobs are added to the queue
- [x] Fix the latest block update logic
- [x] Fix v2 accounts gateway response mapping schema

### How was it tested?
Local + Jenkins


nagdahimanshu (approved):

services/gateway/sources/version2/mappings/account.js
```diff
@@ -69,7 +65,14 @@ module.exports = {
...
-		sentVotes: '=',
-		unlocking: '=',
+		sentVotes: ['sentVotes', {
+			delegateAddress: '=,number',
+			amount: '=,number',
+		}],
+		unlocking: ['unlocking', {
+			delegateAddress: '=,number',
+			amount: '=,number',
+			unvoteHeight: '=,number',
```
nagdahimanshu, line 75: Shouldn't `unvoteHeight` should be an object consists of start and height (as in `services/core/shared/core/compat/sdk_v5/accounts.js`, we are calculating `unvoteHeight` as a object),

```
	unlocking: ['unlocking', {
		amount: '=,string',
		height: {
			start: '=,string',
			end: '=,string',
		},
		delegateAddress: '=,string',
	}],
```
We can clear this with @MichalTuleja ?

sameersubudhi: The change in the mapping is in accordance with the response example in the specification document. In case, it has to be an object we should update the `start` and `end` heights to be instances of `number` instead, for consistency.

```
	unlocking: ['unlocking', {
		amount: '=,string',
		height: {
			start: '=,number',
			end: '=,number',
		},
		delegateAddress: '=,string',
	}],
```

nagdahimanshu: Let's wait for @MichalTuleja reply, what he says?

services/core/shared/core/compat/sdk_v5/blocks.js
```diff
@@ -148,7 +148,12 @@ const isQueryFromIndex = params => {
 	const isDirectQuery = ['id', 'height', 'heightBetween'].some(prop => paramProps.includes(prop));
 
 	const sortOrder = params.sort ? params.sort.split(':')[1] : undefined;
-	const isLatestBlockFetch = (paramProps.length === 3 && params.limit === 1 && params.offset === 0 && sortOrder === 'desc');
+	const isLatestBlockFetch = (paramProps.length === 1 && params.limit === 1)
+		|| (paramProps.length === 2
+			&& ((params.limit === 1 && params.offset === 0)
+				|| (sortOrder === 'desc' && (params.limit === 1 || params.offset === 0))
+			))
+		|| (paramProps.length === 3 && params.limit === 1 && params.offset === 0 && sortOrder === 'desc');
```
MichalTuleja, line 156: Can you please remind me why is it so important to determine whether is it the last block or not?
What `isLatestBlockFetch` literally means? Latest block to fetch?

services/gateway/sources/version2/mappings/account.js
MichalTuleja: According to specification in the issue #250, the property is called `unvoteHeight` and defined as `Number`.

```
      "unlocking": [
        {
          "delegateAddress": "lsk24cd35u4jdq8szo3pnsqe5dsxwrnazyqqqg5eu",
          "amount": "string number",
          "unvoteHeight": 30
        }
      ]
```

Do you think the `Number` might be a problem here?

MichalTuleja (changes requested):
I've added my notes. Btw, the function `indexNewBlocks` is not entirely clear what it does. We've got `indexBlocks` and `indexNewBlocks` so that's quite confusing.

What function is being used to on the new block arrival?

services/gateway/sources/version2/mappings/account.js
nagdahimanshu: I dont think its a problem here as from `lisk-core` we are getting `unvoteHeight` directly.

MichalTuleja: Do you mean we get it as a `Number`? Then we can make it exactly as described above.

nagdahimanshu: Yes, we get `unvoteHeight` as a `Number` directly from core.

services/core/shared/core/compat/sdk_v5/blocks.js
sameersubudhi: `isLatestBlockFetch` evaluates the request params that determine the conditions on which the latest forged block is to be fetched from Lisk Core. If unclear, we could rename it to `isFetchLastForgedBlockFromCore` or to something else. I'm open to recommendations.

Without this check, we would always end up querying our `blocksDB` index. Thus, we would never update the index with the newly forged blocks from Lisk Core. This was the issue because of which our index wasn't getting updated automatically even though `performLastBlockUpdate` was executed everytime the `newBlock` event (signal) was fired.

sameersubudhi:
> I've added my notes. Btw, the function `indexNewBlocks` is not entirely clear what it does. We've got `indexBlocks` and `indexNewBlocks` so that's quite confusing.
> 
> What function is being used on the new block arrival?

`indexBlocks` is the method that gets executed within the job queue and is primarily responsible for writing all the block information to the `blocks` index.

`indexNewBlocks` on the other hand is responsible to determine if a block is already indexed. If not, it queues the new block for indexing and updates the block finality status (`isFinal` flag) for the non-final blocks present on the index, wherever applicable.

services/gateway/sources/version2/mappings/account.js
sameersubudhi: This is exactly what the current change does and is in line with the specifications in #250.

sameersubudhi: As concluded in our last discussion, `unlocking` is now updated to 
```
		unlocking: ['unlocking', {
			delegateAddress: '=,string',
			amount: '=,string',
			height: {
				start: '=,number',
				end: '=,number',
			},
		}],
```

MichalTuleja (approved):

PR #430, opened 2021-04-01
Reclaim transaction support - Closes #417

### LINKED ISSUE #417: Reclaim transaction support
### Description

Accounts migrated to the SDK version 5 blockchain need to support several migration scenarios.

- Accounts that have a public key associated in the legacy blockchain are usable just with a passphrase -> no specific migration is needed from user’s perspective, just sign in with a passphrase is required.
- Accounts that are missing the public key in the legacy blockchain need to reclaim funds through the reclaim transaction. All information needed to make this happen can be retrieved from the genesis block and Lisk Service needs to support that method of migration. 

From the technical point of view, the user that types a correct passphrase of the account that does not exist in the blockchain. In that scenario another request has to be made. 
 - The Lisk Core `getUnregisteredAccount` method helps us determine whether the account has funds that might need to be reclaimed.
 - This feature can be implemented within the `/api/v2/accounts` endpoint, by adding legacy properties.

### Motivation

- Currently only new and genesis-based accounts are retrieved properly by Lisk Service.
- All migration methods need to be supported.

### Acceptance Criteria

- Reclaim transaction properties are retrieved from Lisk Core node.
- Reclaim transaction parameters are passed through the API, so the UI products can use it.

### Additional Information

Make the `/api/v2/account` to be able to search for tokens to reclaim from the legacy blockchain.

The `publicKey` parameter is the only method to retrieve legacy information.
Searching by other account parameters will result in missing legacy information.

Access to the legacy blockchain is not needed, the information is stored in the genesis block.

Extend the response provide the following properties:

```jsonc
// HTTP /api/v2/account?publicKey=968ba2fa993ea9dc27ed740da0daf49eddd740dbd7cb1cb4fc5db3a20baf341b

{
  "data": {
    "summary": {
      "address": "02a1de92405edeaac13913bd089b07eb73cbd762",
      "legacyAddress": "2841524825665420181L",
      "balance": "0" // balance in the new blockchain
      "isMigrated": true, // shows only when found by the public key
                          // true = account is migrated from the legacy blockchain
                          // false = account is created in the current blockchain
      "publicKey": "968ba2fa993ea9dc27ed740da0daf49eddd740dbd7cb1cb4fc5db3a20baf341b",
      "isDelegate": false,
      "isMultisignature": false,
    },
    "token": {
      "balance": "0" // balance in the new blockchain
    },
    "sequence": { ... },
    "keys": { ... },
    "dpos": { ... },
    "legacy": { // only if the reclaim wasn't made yet
      "address": "2841524825665420181L", // Legacy address in hex format
      "balance": "151146419900", // balance to reclaim
    },
  },
  "meta": {
    "count": 1,
    "offset": 0
  },
  "links": {}
}
```

sameersubudhi:
### What was the problem?
This PR resolves #417 

### How was it solved?
- [x] Wrap MySQL batch inserts within a transaction
- [x] Fix genesis account filtering logic to filter out the legacy addresses
- [x] Wait for the account indexing job to be queued
- [x] Separate sets of controllers differentiated by API version
- [x] Update gateway mappings to include `isMigrated`, `legacyAddress` and `legacy` information in response
- [x] Update document with the corresponding change

### How was it tested?
Local


MichalTuleja (approved):

nagdahimanshu (approved):
LGTM 👍

PR #436, opened 2021-04-06
Fix transaction statistics - Closes #433

### LINKED ISSUE #433: Fix transaction statistics
### Transaction statistics are not calculated properly 

```2021-04-01T20:53:45.929 WARN [queue] transactionStatisticsQueue Job failed,TypeError: Cannot assign to read only property '0' of string 'Invalid Number'```

### Expected behavior
Stats calculated properly with no error

### Steps to reproduce
Run Lisk Service against lisk-core betanet5 and call `api/v2/transactions/statistics/day`

### PR DESCRIPTION
**Pull request by nagdahimanshu**
### What was the problem?
This PR resolves #433 

### How was it solved?
- [x] Fix transaction stats issues
  - [x] Fix for calculating stats properly
  - [x] Fix for incorrect `transactionCount` issue
  - [x] Verified changes against: `betanet5`, `betanet` and `testnet`
- [x] Increase pool size for DB timeout
- [x] Fix account/delegate gateway mapping to include `sentVotes` & `unlocking`
- [x] Support for special charset mysql

### How was it tested?
- [x] Local
- [x] Against betanet5
- [x] Jenkins

### DIFF
--- a/services/core/shared/core/transactionStatistics.js
+++ b/services/core/shared/core/transactionStatistics.js
@@ -22,6 +22,7 @@ const { getTransactions } = require('./transactions');
 const { initializeQueue } = require('./queue');
 const mysql = require('../indexdb/mysql');
 const signals = require('../signals');
+const requestAll = require('../requestAll');
 
 const logger = Logger();
 
@@ -142,8 +143,8 @@ const fetchTransactions = async (date, offset = 0) => {
 		limit,
 		offset,
 	};
-	const transactions = await getTransactions(params);
-	return transactions.data;
+	const transactions = await requestAll(getTransactions, params, 20000);
+	return transactions;
 };
 
 const queueJob = async (job) => {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: services/core/shared/core/transactionStatistics.js [hunk 1]
```diff
@@ -142,8 +143,8 @@ const fetchTransactions = async (date, offset = 0) => {
 		limit,
 		offset,
 	};
-	const transactions = await getTransactions(params);
-	return transactions.data;
+	const transactions = await requestAll(getTransactions, params, 20000);
```
sameersubudhi, line 146: Instead of hardcoding the limit value at 20000, it might be better to invoke `getTransactions` with `params` (but with `limit: 1`, eg: `{ ...params, limit: 1 }`) and pass `meta.total` from the response to `requestAll` as the limit.

sameersubudhi (changes requested):

services/core/shared/core/transactionStatistics.js
nagdahimanshu: Done!

sameersubudhi (approved):
LGTM! Works well 👍

MichalTuleja (approved):
Thanks for these contributions 👍

PR #487, opened 2021-04-28
Jenkins test

MichalTuleja:
Simple jenkins test.


tests/integration/api_v2/http/blocks.test.js
```diff
@@ -43,7 +43,7 @@ describe('Blocks API', () => {
 		[refBlock] = (await api.get(`${endpoint}?limit=1&offset=2`)).data;
 		do {
 			// eslint-disable-next-line no-await-in-loop
-			response = await api.get(`${baseUrlV2}/accounts?isDelegate=true&limit=1`);
+			response = await api.get(`${baseUrlV2}/accounts?isDelegate=true&limit=1&search=genesis`);
```
MichalTuleja, line 46: @sameersubudhi Why is this added here? Is it required to make the tests suite passing?

sameersubudhi: @MichalTuleja The test blockchain has a non-forging delegate that we registered by the name of `test_delegate`, which when used as a reference delegate results in the failing test cases (fetch blocks by the reference delegate properties).
Hence, to deterministically fetch a forging delegate this change is necessary. The premise is that all the `genesis_` delegates on the test blockchain are forging by default.

PR #497, opened 2021-05-04
Improve integration tests - Closes #445

sameersubudhi:
### What was the problem?
This PR resolves #445 

### How was it solved?
- [ ] TODO

### How was it tested?
Local + Jenkins

PR #500, opened 2021-05-04
Sorting by amount returns null values - Closes #499

### LINKED ISSUE #499: Sorting by amount returns null values
### Actual behavior

Sorting by amount returns null values.

### Expected behavior

Sorting by amount returns only non-null values.

### Steps to reproduce

Retrieve transactions sorted by amounts.

sameersubudhi:
### What was the problem?
This PR resolves #499 

### How was it solved?
- [x] Perform sorting only on non-null values
- [x] Merge sorting based test cases from PR https://github.com/LiskHQ/lisk-service/pull/497
- [x] Enable disabled transactions test cases 
- [x] Ensure `startHeight` is not lower than `genesisHeight` when checking for missing blocks

### How was it tested?
Local + Jenkins

### DIFF
--- a/services/core/shared/indexdb/mysql.js
+++ b/services/core/shared/indexdb/mysql.js
@@ -271,15 +271,6 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
 		const query = knex.count(`${tableConfig.primaryKey} as count`).table(tableName);
 		const queryParams = resolveQueryParams(params);
 
-		if (params.orWhere) {
-			const { orWhere } = params;
-			query.where(function () {
-				this.where(queryParams).orWhere(orWhere);
-			});
-		} else {
-			query.where(queryParams);
-		}
-
 		if (params.propBetweens) {
 			const { propBetweens } = params;
 			propBetweens.forEach(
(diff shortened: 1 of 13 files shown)


services/core/shared/indexdb/mysql.js
```diff
@@ -214,6 +214,7 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
 		if (params.sort) {
 			const [sortProp, sortOrder] = params.sort.split(':');
 			query.orderBy(sortProp, sortOrder);
+			query.whereNotNull(sortProp);
```
nagdahimanshu, line 217: It works well but returns incorrect count. Now, we need to add given logic to our count function as well to get exact count: 
```
if (params.sort) {
  const [sortProp, sortOrder] = params.sort.split(':');
  query.orderBy(sortProp, sortOrder);
  query.whereNotNull(sortProp);
}
```

nagdahimanshu (changes requested):
Please consider one suggestion!

nagdahimanshu (approved):
LGTM!

MichalTuleja (approved):
LGTM 👍

PR #543, opened 2021-05-28
Provide votes as aggregated values - Closes #486

### LINKED ISSUE #486: Provide votes as aggregated values
### Description

The vote amounts returned from the on `/api/v2/votes_received` endpoint should be aggregated values, instead of incremental values as transactions are done by users over time.

### Motivation

- Provide lisk desktop with the data that it needs for its UI without having to do expensive calculation itself
- Reduce clutter. Consumers of this endpoint would probably be only interested in aggregated values

### Acceptance Criteria

- Provide vote amounts as aggregated values

### Additional information

Add the `aggregate` param that takes boolean value. Keep it `false` by default.

Use database incrementation for newly arrived blocks.

```jsonc
// /api/v2/votes_received?address=lsk24cd35u4jdq8szo3pnsqe5dsxwrnazyqqqg5eu&aggregate=true // default is false
{
  "data": {
    "account": {
      "address": "lsk24cd35u4jdq8szo3pnsqe5dsxwrnazyqqqg5eu",
      "username": "genesis_56",
      "votesReceived": 10
    },
    "votes": [ // summary of votes
      {
        "address": "lsks6uckwnap7s72ov3edddwgxab5e89t6uy8gjt6",
        "amount": 1081560729258, // sum of all votes per account
        "username": "liskhq"
      }
    ]
  },
  "meta": {
    "count": 10,
    "offset": 0,
    "total": 10
  },
  "links": {}
}
```

```jsonc
// /api/v2/votes_received?address=lsk24cd35u4jdq8szo3pnsqe5dsxwrnazyqqqg5eu&aggregate=false
{
  "data": {
    "account": {
      "address": "lsk24cd35u4jdq8szo3pnsqe5dsxwrnazyqqqg5eu",
      "username": "genesis_56",
      "votesReceived": 10
    },
    "votes": [ // list of particular votes
      {
        "address": "lsks6uckwnap7s72ov3edddwgxab5e89t6uy8gjt6",
        "amount": 1081560729258,
        "username": "liskhq"
      },
      {
        "address": "lsks6uckwnap7s72ov3edddwgxab5e89t6uy8gjt6",
        "amount": 1081560729258,
        "username": "liskhq"
      }
    ],
  },
  "meta": {
    "count": 10,
    "offset": 0,
    "total": 10
  },
  "links": {}
}
```

nagdahimanshu:
### What was the problem?
This PR resolves #486 

### How was it solved?
- [x] Update gateway:
  - [x] Add new request param `aggregate` to `/votes_received`
  - [x] Set `false` as default value for `aggregate`
  - [x] Update `votesUsed ` to `votesReceived`
- [x] Update core:
  - [x] Add new schema param `votesAggregated`
  - [x] Fix typo in voters schema
  - [x] Update `indexVotes` to support incrementing vote amount on every new vote transaction
  - [x] Update `mysql/increment` method to support `insert if not exists` or `increment if exists`
  - [x] Update `getVoters` method to support returning `aggregated votes per account` when aggregate=true
- [x] Fix `votes_received` bug (returns incorrect votes per account)
- [x] Update `rewards/producedBlocks` logic to insert if not exists or increment

### How was it tested?
- [x] Local
- [x] Jenkins


services/core/shared/indexdb/mysql.js
```diff
@@ -318,12 +318,23 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
...
+			);
+			return result;
+		} catch (error) {
+			return knex.transaction(
+				trx => trx(tableName)
+					.where(params.where.property, '=', params.where.value)
+					.increment(params.increment)
+					.transacting(trx),
+			);
+		}
```
sameersubudhi, line 336: Please consider replacing the `try...catch` with a simpler `if..else` conditional.
Also, could you please confirm if the return values in both the cases are consistent?
```suggestion
		if (row) {
			const [result] = await knex.transaction(
				trx => trx(tableName)
					.insert(row)
					.transacting(trx),
			);
			return result;
		} else {
			return knex.transaction(
				trx => trx(tableName)
					.where(params.where.property, '=', params.where.value)
					.increment(params.increment)
					.transacting(trx),
			);
		}
```

services/core/shared/core/compat/sdk_v5/voters.js
```diff
@@ -34,22 +34,39 @@ const extractAddressFromPublicKey = pk => (getAddressFromPublicKey(Buffer.from(p
...
+							amount: BigInt(vote.amount),
+						},
+						where: {
+							property: 'id',
+							value: voteEntry.receivedAddress.concat(voteEntry.sentAddress),
+						},
+					}, {
+						...voteEntry,
+						id: voteEntry.receivedAddress.concat(voteEntry.sentAddress),
+					});
```
sameersubudhi, line 64: Kindly consider adding an additional check to verify that the transaction containing the vote information isn't already indexed before performing any increment. This is necessary to avoid replays.

Alternatively, it might also be beneficial to add a flag within the original `votes` table that tracks if a particular vote transaction is already reflected within the `votes_aggregated` index.

services/core/shared/core/compat/sdk_v5/voters.js
```diff
@@ -104,23 +120,13 @@ const getVoters = async params => {
...
-			voteMultiArray
-				.forEach(lvotes => allVotes = allVotes.concat(lvotes.map(v => ({
-					...normalizeVote(v),
-					sentAddress: getBase32AddressFromHex(extractAddressFromPublicKey(v.senderPublicKey)),
-				}))));
-			votes.data.votes = allVotes;
-		}
+	let resultSet;
+	if (params.aggregate) {
+		resultSet = await votesAggregateDB.find({ sort: 'timestamp:desc', receivedAddress: params.receivedAddress });
```
sameersubudhi, line 125: The logic currently does not support `limit` and `offset`. Please make the necessary changes.

services/core/shared/core/compat/sdk_v5/voters.js
```diff
@@ -137,10 +143,10 @@ const getVoters = async params => {
 	votes.data.account = {
 		address: params.receivedAddress,
 		username: accountInfo && accountInfo.username ? accountInfo.username : undefined,
-		votesUsed: votes.data.votes.length,
+		votesReceived: votes.data.votes.length,
```
sameersubudhi, line 146: After including the support for `limit` and `offset`, `votesReceived` can be updated with `votesAggregateDB.count({.receivedAddress: params.receivedAddress })`.

sameersubudhi (changes requested):
Please find my observations below.

services/core/shared/indexdb/mysql.js
nagdahimanshu: It has been clarified offline and response is consistent now.

services/core/shared/core/compat/sdk_v5/voters.js
nagdahimanshu: It already support `limit` and `offset`.

services/core/shared/core/compat/sdk_v5/voters.js
nagdahimanshu: Done!

services/core/shared/core/compat/sdk_v5/voters.js
nagdahimanshu: Added additional logic to verify that.

sameersubudhi (approved):
LGTM 👍

services/core/shared/indexdb/mysql.js
```diff
@@ -318,12 +318,26 @@ const getDbInstance = async (tableName, tableConfig, connEndpoint = config.endpo
...
+			);
+			// always return number of rows affected
+			result = result === 0 ? 1 : 0;
+		} catch (error) {
+			result = await knex.transaction(
+				trx => trx(tableName)
+					.where(params.where.property, '=', params.where.value)
+					.increment(params.increment)
+					.transacting(trx),
+			);
```
MichalTuleja, line 337: What happens when that transaction throws an error as well?

services/core/shared/core/compat/sdk_v5/schema/votesAggregate.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+module.exports = {
+	primaryKey: 'id',
+	schema: {
+		id: { type: 'string' },
+		sentAddress: { type: 'string' },
+		receivedAddress: { type: 'string' },
+		amount: { type: 'bigInteger', defaultValue: 0 },
+		timestamp: { type: 'integer' },
```
MichalTuleja, line 23: What timestamp is presented here? If it's last update it has to be mentioned in the docs.

nagdahimanshu: Yes, its last update time (last vote received to given account from any account). It's only for internal purpose and we dont send it in response. Also, I have update the docs to include `aggregate` params for `votes_received` now.

services/core/shared/indexdb/mysql.js
nagdahimanshu: We are using job `queue` for `indexVotes`. So, If it fails, queue will retry again.

MichalTuleja (approved):

PR #561, opened 2021-06-04
Fix delegate properties - Closes #550

### LINKED ISSUE #550: Fix delegate properties
### Actual behavior

The `lastForgedHeight` property returns delegate registration height if there are no forged blocks by this delegate.

### Expected behavior

1. Make it distinguish the `lastForgedHeight` from the registration height.
2. In case the delegate never forged any block the `lastForgedHeight` has to be `null`.
3. Add the `registrationHeight`property to the API forging summary.
4. Make `producedBlocks` and `rewards` when the account did never forge.

```jsonc
{
    "data": [
        {
            "dpos": {
                "delegate": {
                    "consecutiveMissedBlocks": 0,
                    "isBanned": false,
                    "lastForgedHeight": 984987, // this should be null
                    "producedBlocks": null, // this should be 0
                    "rewards": null,         // this should be 0
                    "registrationHeight": 984987, // new property
                    "totalVotesReceived": "0",
                    "username": "sri"
                },
...
```

### Steps to reproduce

Call the URL
https://betanet5-service.lisk.io/api/v2/accounts?username=sri

`0.3.0`

sameersubudhi:
### What was the problem?
This PR resolves #550 

### How was it solved?
- [x] `producedBlocks` is set to `0` for non-forging delegate accounts
- [x]  `rewards` is set to `0` for non-forging delegate accounts
- [x] `registrationHeight` is now returned for delegate accounts
- [x] `lastForgedHeight` is set to `null` for non-forging delegates
- [x] `lastForgedHeight` is set based on the `blocks` index
- [x] `registrationHeight` is returned only after the complete blockchain is successfully indexed
- [x] `registrationHeight` for a delegate defaults to the genesis block height, if no delegate registration transaction is available in the index
- [x] Set higher preference for a direct `targetPair` match from the source prices over calculated rates

### How was it tested?
Local + Jenkins

### DIFF
--- a/services/core/shared/core/compat/sdk_v5/accounts.js
+++ b/services/core/shared/core/compat/sdk_v5/accounts.js
@@ -46,11 +46,13 @@ const coreApi = require('./coreApi');
 const mysqlIndex = require('../../../indexdb/mysql');
 
 const accountsIndexSchema = require('./schema/accounts');
+const blocksIndexSchema = require('./schema/blocks');
 const transactionsIndexSchema = require('./schema/transactions');
 
 const { ValidationException } = require('../../../exceptions');
 
 const getAccountsIndex = () => mysqlIndex('accounts', accountsIndexSchema);
+const getBlocksIndex = () => mysqlIndex('blocks', blocksIndexSchema);
 const getTransactionsIndex = () => mysqlIndex('transactions', transactionsIndexSchema);
 
 // A boolean mapping against the genesis account addresses to indicate migration status
@@ -146,6 +148,10 @@ const resolveDelegateInfo = async accounts => {
 		accounts,
 		async account => {
 			if (account.isDelegate) {
+				const blocksDB = await getBlocksIndex();
+				const transactionsDB = await getTransactionsIndex();
+				const delegateRegTxModuleAssetId = '5:0';
+
 				account.account = {
 					address: account.address,
 					publicKey: account.publicKey,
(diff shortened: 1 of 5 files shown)


MichalTuleja (changes requested):
Looks good but forging delegates resolve to `"registrationHeight": null`.
Is it possible to make this property filled for forging delegates?

sameersubudhi:
> Looks good but forging delegates resolve to `"registrationHeight": null`.
> Is it possible to make this property filled for forging delegates?

`registrationHeight` is returned for all the delegates, once all the transaction index is synced completely. It returns `null` until the delegate registration transaction is not indexed. Alternatively, I could return the property only after the indexes are fully synchronized.

services/core/shared/core/compat/sdk_v5/accounts.js
```diff
@@ -176,6 +175,32 @@ const resolveDelegateInfo = async accounts => {
...
 					.sort((a, b) => b - a).slice(0, 5)
 					.map(height => ({ start: height, end: height + punishmentHeight }));
+
+				const [lastForgedBlock = {}] = await blocksDB.find({
+					generatorPublicKey: account.publicKey,
+					limit: 1,
+				});
+				account.dpos.delegate.lastForgedHeight = lastForgedBlock.height || null;
+
+				// Iff the COMPLETE blockchain is SUCCESSFULLY indexed
```
nagdahimanshu, line 185:
```suggestion
				// If the COMPLETE blockchain is SUCCESSFULLY indexed
```

nagdahimanshu (approved):
Looks good apart from one typo 👍

services/core/shared/core/compat/sdk_v5/accounts.js
sameersubudhi: @nagdahimanshu It is not a typo.`iff` is the acronym for `if and only if`.

nagdahimanshu: Thanks for clarification 👍

MichalTuleja (approved):

PR #593, opened 2021-06-17
Migrate newsfeed to framework - Closes #577

### LINKED ISSUE #577: Migrate news feed service to the framework
### Description

Update the solution to leverage the lisk-service-framework. 

This step consists of updating the app.js and creating the basic directory structure.

Make sure that files like `Dockerfile` and `package.json` are up-to-date.

### Motivation

To achieve better maintainability

### Acceptance Criteria

- The application should be able to run after that step

MichalTuleja:
### What was the problem?

This PR resolves #577

### How was it solved?

- [x] Updated `package.json`
- [x] Migrated logger to framework
- [x] Migrated mapper to framework
- [x] Migrated app.js to framework
- [x] Removed unused files after framework migrations
- [x] Removed unused imports from `package.json`
- [x] Made app.js running
- [x] Moved configs
- [x] Updated `Dockerfile`
- [x] Updated configs

### How was it tested?

Manually


services/newsfeed/shared/drupal.js
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+module.exports = {};
```
sameersubudhi, line 16: Please consider renaming the `shared` directory to `sources` instead.

services/newsfeed/shared/twitter.js
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+const Twitter = require('twitter');
```
sameersubudhi, line 16: Same as above.

services/newsfeed/config_drupal.js
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+const getDrupalConfig = ({ enabled, name, url, filter }) => ({
```
sameersubudhi, line 16: Since these are source-specific configs, kindly consider moving/renaming it to `services/newsfeed/sources/config/drupal.js`

services/newsfeed/config_twitter.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+const getTwitterConfig = () => ({
```
sameersubudhi, line 16: Same as above: `services/newsfeed/sources/config/twitter.js`

sameersubudhi (changes requested):

services/newsfeed/config_drupal.js
MichalTuleja: This file won't stay here for long. Add Drupal data source #582 will be a good issue to make it sorted, same with twitter.

services/newsfeed/shared/drupal.js
MichalTuleja: I was planning on keeping the same dir structure (jobs & methods use shared files). Is there any reason to change it? Esp. when there are only 3-5 files?

sameersubudhi: Then to keep the directory structure consistent with the other microservices should we not move these files to `shared/newsfeed`?
The files maintained directly under the `shared` directory (within `core` microservice) are highly generic files and not directly related to the API endpoints.

MichalTuleja: The `core` microservice needs directories for two reasons: 1) the number of files is already high 2) some files got the same names and might require long prefixes otherwise. None of them is the case here. When the scope becomes bigger, the files will be organised into directories.

nagdahimanshu (approved):

services/newsfeed/methods/controllers/newsfeed.js
```diff
@@ -0,0 +1,41 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+const sources = require('../../config.sources.js');
+
+const enabledSources = Object.values(sources)
```
sameersubudhi, line 18:
```suggestion
const config = require('../../config.js');

const enabledSources = Object.values(config.sources)
```

sameersubudhi (approved):
LGTM 👍

PR #602, opened 2021-06-21
Configurable genesis height - Closes #601

### LINKED ISSUE #601: Configurable genesis height
### Description

The default genesis height in Lisk Service is set to `0`.

This is the arbitrary setting that is true on most development-related networks.

Make genesis height configurable through the environment variable `GENESIS_HEIGHT`.

Set it by default to `0`.

### Motivation

- Migrated networks are not able to benefit Lisk Service with hardcoded genesis height.

### Acceptance Criteria

- Lisk Service works properly with migrated networks.

MichalTuleja:
### What was the problem?

This PR resolves #601

### How was it solved?

- [x] Added `GENESIS_HEIGHT` variable
- [x] Test blockchain index
- [x] Test transaction statistics
- [x] Test fee estimator

### How was it tested?

Jenkins

### DIFF
--- a/services/core/shared/core/compat/sdk_v5/dynamicFees.js
+++ b/services/core/shared/core/compat/sdk_v5/dynamicFees.js
@@ -142,7 +142,8 @@ const getEstimateFeeByteForBlock = async (blockBatch, innerPrevFeeEstPerByte) =>
 
 const getEstimateFeeByteForBatch = async (fromHeight, toHeight, cacheKey) => {
 	const transactionsDB = await getTransactionsIndex();
-	const { defaultStartBlockHeight, genesisHeight } = config.feeEstimates;
+	const { genesisHeight } = config;
+	const { defaultStartBlockHeight } = config.feeEstimates;
 
 	// Use incrementation to skip the genesis block - it is not needed
 	const calculationHeight = Math.max(defaultStartBlockHeight, genesisHeight + 1);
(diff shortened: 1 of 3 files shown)


services/core/shared/core/compat/sdk_v5/dynamicFees.js
```diff
@@ -142,10 +144,13 @@ const getEstimateFeeByteForBlock = async (blockBatch, innerPrevFeeEstPerByte) =>
...
 	const transactionsDB = await getTransactionsIndex();
+	const { genesisHeight } = config;
+	const { defaultStartBlockHeight } = config.feeEstimates;
 
 	// Check if the starting height is permitted by config or adjust acc.
-	fromHeight = config.feeEstimates.defaultStartBlockHeight > fromHeight
-		? config.feeEstimates.defaultStartBlockHeight : fromHeight;
+	// Use incrementation to skip the genesis block - it is not needed
+	fromHeight = Math.max(...[defaultStartBlockHeight, genesisHeight + 1, fromHeight]
+		.filter(n => !Number.isNaN(n)));
```
sameersubudhi, line 153: Kindly consider moving the logic to the shared layer instead, so that it is available for all SDK versions.

sameersubudhi (changes requested):

nagdahimanshu (approved):
LGTM!

services/core/shared/core/compat/sdk_v5/dynamicFees.js
MichalTuleja: That logic doesn't seem to be specific to all SDK versions. Can you explain how would you like to apply that remark?
Another thing, how do you test the patch against the SDKv4 when there is no migrator for prior versions?

sameersubudhi: Please ignore the comment. The change seems unnecessary.

sameersubudhi (approved):
LGTM!

PR #634, opened 2021-07-12
Add support for HTTP 'Cache-Control' - Closes #632

### LINKED ISSUE #632: Add support for HTTP 'Cache-Control'
### Description

In order to support caching at the reverse-proxy, the HTTP `Cache-Control` is necessary for the response header.

Make sure the gateway adds it to the response headers.

### Motivation

Caching eliminates the unnecessary repetitive API invocation of the same endpoint URL for which the response is expected not to vary for a certain duration.

### Additional Information

Acceptance criteria: `Cache-Control` header with the `public` and `max-age` directives are added to the response.

sameersubudhi:
### What was the problem?
This PR resolves #632 

### How was it solved?
- [x] Add `Cache-Control` header to the API response
- [x] Add switch to enable (disabled by default) the header in the response
- [x] Add environment variable to allow customization of the `Cache-Control` directives
- [x] Update docker environment file to include the above variables

### How was it tested?
Local

### DIFF
--- a/docker/lisk-service/env/gateway.env
+++ b/docker/lisk-service/env/gateway.env
@@ -5,3 +5,7 @@ HOST=0.0.0.0
 # Enabled APIs
 ENABLE_HTTP_API=http-version1,http-version1-compat,http-status,http-version2
 ENABLE_WS_API=rpc,rpc-v1,blockchain,rpc-v2
+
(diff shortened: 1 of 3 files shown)


nagdahimanshu (approved):
LGTM 👍

PR #639, opened 2021-07-16
Fix delegate rank calculation - Closes #638

### LINKED ISSUE #638: Fix incorrect delegate rank calculation
### Actual behavior 
Delegate property 'rank' is not calculated properly

### Expected behavior
Delegate rank should be calculated on every new block and return correct delegate list

### Steps to reproduce
Run Lisk Service on local and check the rank of delegates

0.4.0

nagdahimanshu:
### What was the problem?
This PR resolves #638 

### How was it solved?
- [x] Implemented logic to update only specific delegates which get updated on every block (instead of updating the whole list)
- [x] Update cron job to schedule `reloadDelegatesCache` in every 5 minutes
- [x] Add sleep in Jenkinsfile
- [x] Support for special character `MySQL` in core service
 
### How was it tested?
Local + Jenkins

### DIFF
--- a/services/core/shared/core/delegates.js
+++ b/services/core/shared/core/delegates.js
@@ -13,7 +13,7 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-const { Logger, CacheRedis } = require('lisk-service-framework');
+const { Logger, CacheRedis, Signals } = require('lisk-service-framework');
 const BluebirdPromise = require('bluebird');
 
 const config = require('../../config');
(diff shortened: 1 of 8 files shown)


services/core/shared/core/delegates.js
```diff
@@ -249,6 +249,34 @@ const reload = async () => {
...
+			if (indexOfDelegate === -1) {
+				delegateList.push(delegate);
+			} else {
+				delegateList[indexOfDelegate] = delegate;
+			}
+		});
+		await computeDelegateRank();
+		await computeDelegateStatus();
+	}
+});
```
sameersubudhi, line 278: Delegate `rank` would vary only when the `voteWeight` changes and the `status` gets updated at the beginning of a new round. 

```suggestion
// Keep the delegate cache up-to-date
Signals.get('newBlock').add(async data => {
	const dposModuleId = 5;
	const voteDelegateAssetId = 1;

	const updatedDelegateAddresses = [];

	const [block] = data.data;
	if (block && block.payload) {
		block.payload.forEach(tx => {
			if (tx.moduleID === dposModuleId && tx.assetID === voteDelegateAssetId) {
				tx.asset.votes.forEach(vote => updatedDelegateAddresses
					.push(coreApi.getBase32AddressFromHex(vote.delegateAddress)));
			}
		});

		const { data: updatedDelegateAccounts } = await coreApi
			.getAccounts({ addresses: updatedDelegateAddresses });

		updatedDelegateAccounts.forEach(delegate => {
			const delegateIndex = delegateList.findIndex(acc => acc.address === delegate.address);
			if (delegateIndex === -1) delegateList.push(delegate);
			else delegateList[delegateIndex] = delegate;
		});

		// Rank is impacted only when a delegate gets (un-)voted
		if (updatedDelegateAddresses.length) await computeDelegateRank();
	}
});
```

sameersubudhi (changes requested):
Please find my review.

MichalTuleja (approved):

sameersubudhi (approved):
LGTM 👍

PR #649, opened 2021-07-21
Update Signal listener - Closes # 648

### LINKED ISSUE #648: Update Signal listener
### Description

The function `Signal.has(listener)` should be used to verify before adding a new listener `Signal.add(listener)`.

Ref: https://github.com/millermedeiros/js-signals/blob/1a82bdd4097941bb010ab4a5757a67da460ce2f4/dist/signals.js#L285

### Motivation

- Signal listeners are essential part of the application
- They need to be reliable

sameersubudhi:
### What was the problem?
This PR resolves #648 

### How was it solved?
- [x] Prevent adding a listener to the Signal if it already is registered
- [x] Simplify implementation for `getApiClient`
- [x] Ensure `isFeeEstimatesReady` is not ready when the response status is `SERVICE_UNAVAILABLE`
- [x] Consider the highest finalized indexed block height to continue indexing the past blocks (cherry-picked from #663)

### How was it tested?
Local

### DIFF
--- a/services/core/events/blockchain.js
+++ b/services/core/events/blockchain.js
@@ -28,7 +28,7 @@ module.exports = [
 		name: 'block.change',
 		description: 'Keep the block list up-to-date',
 		controller: callback => {
-			Signals.get('newBlock').add(async data => {
+			const newBlockListener = async data => {
 				if (data && Array.isArray(data.data)) {
 					const [block] = data.data;
 					logger.debug(`New block arrived (${block.id})...`);
@@ -47,14 +47,15 @@ module.exports = [
 						util.inspect(data),
 					].join('\n'));
 				}
-			});
+			};
+			Signals.get('newBlock').add(newBlockListener);
 		},
 	},
 	{
 		name: 'transactions.new',
 		description: 'Keep newly added transactions list up-to-date',
 		controller: callback => {
-			Signals.get('newBlock').add(async (data) => {
+			const newTransactionsListener = async (data) => {
 				if (data && Array.isArray(data.data)) {
 					const [block] = data.data;
 					if (block.numberOfTransactions > 0) {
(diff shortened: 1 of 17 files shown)


services/core/events/blockchain.js
```diff
@@ -49,14 +49,19 @@ module.exports = [
 						util.inspect(data),
 					].join('\n'));
 				}
-			});
+			};
+			if (!Signals.get('newBlock').has(newBlockListener)) {
+				logger.debug('newBlock has newBlockListener: ', Signals.get('newBlock').has(newBlockListener));
+				Signals.get('newBlock').add(newBlockListener);
+				logger.debug('newBlock has now registered newBlockListener: ', Signals.get('newBlock').has(newBlockListener));
+			}
```
MichalTuleja, line 57: Is there any reason to keep that logic here instead of implementing in `Signal.get()`?

Please update it if possible.

MichalTuleja (changes requested):

services/core/events/blockchain.js
sameersubudhi: Done!

services/core/events/blockchain.js
```diff
@@ -49,14 +47,15 @@ module.exports = [
 						util.inspect(data),
 					].join('\n'));
 				}
-			});
+			};
+			Signals.get('newBlock').add(newBlockListener);
```
nagdahimanshu, line 51: Is there any reason to wrap the given logic in function before calling? (Ideally, both logic seems to be same)

sameersubudhi: Separate instances of similar anonymous or named functions within Javascript when strictly compared (`===`) evaluate to `false`. Therefore, for `Signals.has()` to evaluate to `true`, we need named listener functions. `has()` internally performs a strict comparison to evaluate the registration of a listener function.

nagdahimanshu (approved):
Looks good 👍

MichalTuleja (approved):
Looks good, thanks @sameersubudhi !

PR #663, opened 2021-07-26
Service Core minor issues - Closes #662

### LINKED ISSUE #662: Patch release Service Core updates
### Description

Make sure the minor details work properly in the Core microservice.

- [x] Update genesis block heights in the config
- [x] Replace mainnet network ID to string `update_after_migration`
- [x] Fix: `generatorUsername` returns `null` on genesis block
- [x] Fix: Transaction statistics should run only if the index is built - prevent from running during start
- [x] Fix: Something causes 100% CPU utilization and the process is killed (mainnet with no transactions)
- [x] Fix: Re-building the index causes `indexBlocksQueue Job failed,TimeoutException: Request timed out when calling 'getAccountsByAddresses' for addresses: 84a16857811838e3749eb5f20cd1d064604472a8,`
- [x] Fix: Genesis block is not being downloaded during start
- [x] Replace `existsSync` and `mkdirSync` with their asynchronous counterparts

### Motivation

- Minor details need to be fixed before the final release

sameersubudhi:
### What was the problem?
This PR resolves #662 

### How was it solved?
- [x] Return `generatorAddress` and `generatorUsername` to be `null`, when they do not exist
- [x] Consider the highest `finalized` indexed block height to continue indexing the past blocks
- [x] Replace mainnet network ID to string `update_after_migration`
- [x] Update default Mainnet genesis height
- [x] Add default config support for other networks (apart from `mainnet` and `testnet`)
- [x] Cherry-pick commits for blocks filtering
- [x] Replace `existsSync` and `mkdirSync` with their asynchronous counterparts
- [x] Use Redis to cache the genesis accounts flag
- [x] Implement retry within `indexMissingBlocks` instead of `buildIndex`
- [x] Reintroduce `waitForIt` based `getApiClient` implementation
- [x] Replace all 'getAccountsByAddresses' invocations with equivalent 'getAccountByAddress'
- [x] Download genesis block at start
- [x] Index delegate accounts followed by blocks and then all the genesis accounts
- [x] Index genesis accounts in batches of `10000` on every new block
- [x] Memory optimization of the JSON parse utility
- [x] Extensive logging
- [x] More error handling

### How was it tested?
Local


nagdahimanshu (approved):
LGTM 👍

MichalTuleja (approved):

MichalTuleja:
Lots of changes 👍  Thanks for those contributions @sameersubudhi  @nagdahimanshu 🎉

PR #670, opened 2021-08-04
Fix delegate vote weight calculation - Closes #669

### LINKED ISSUE #669: Incorrect vote weight calculation for delegates
### Actual behavior

The `totalVotes` computation for a delegate is incorrect and uses `sentVotes` for calculation. 
Eg: The delegate with username `nerigal` ranks `#8` on the testnet.

### Expected behavior

The `totalVotes` computation for a delegate is unnecessary and should be using `totalVotesReceived` instead.
Eg: The delegate with username `nerigal` should rank lower (`#50`).

### Steps to reproduce

https://testnet-service-staging.lisk.io/api/v2/accounts?isDelegate=true&username=nerigal

`v0.4.2`

sameersubudhi:
### What was the problem?
This PR resolves #669 

### How was it solved?
- [x] Use `totalVotesReceived` instead of aggregating `account.dpos.sentVotes` as `totalVotes`
- [x] Pass through the calculated `voteWeight` within the response
- [x] Update revised mainnet genesis block height

### How was it tested?
Local

### DIFF
--- a/docs/api/version2.md
+++ b/docs/api/version2.md
@@ -141,6 +141,7 @@ Make the version 2 API able to retrieve data by those criteria.
         "consecutiveMissedBlocks": 0,
         "lastForgedHeight": 68115,
         "isBanned": false,
+        "voteWeight": "201000000000",
         "totalVotesReceived": "201000000000",
       },
       "sentVotes": [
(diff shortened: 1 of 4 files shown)


nagdahimanshu (approved):
Looks good and works as expected 👍

PR #678, opened 2021-08-05
Legacy account optimisation - Closes #677

### LINKED ISSUE #677: Legacy account optimisation
### Description

Optimize the legacy account retrieval so it avoids additional requests to Lisk Core.

Consider caching the genesis block storing the old address and its balance.

### Acceptance Criteria

- Number of simultaneous requests is reduced
- Caching is implemented (if needed)
- Core timeouts are no longer an issue during account retrieval

sameersubudhi:
### What was the problem?
This PR resolves #677 

### How was it solved?
- [x] Reduce the number of simultaneous requests made to the Core
- [x] Cache legacy account information at init
- [x] Check for reclaim transaction before querying core for legacy information
- [x] Index genesis accounts in even smaller batches
- [x] Core timeouts are no longer an issue during account retrieval
- [x] Make the Gateway service available on all the host machine IPs
- [x] Query transactions in smaller batch sizes (10 at a time)
- [x] Check if `apiClient._channel.invoke` is available before returning the `apiClient` instance
- [x] Set default `brokerTimeout` to `10 sec` for all microservices

### How was it tested?
Local + Jenkins


services/core/shared/core/compat/sdk_v5/accounts.js
```diff
@@ -264,44 +270,70 @@ const indexAccountsbyPublicKey = async (accountInfoArray) => {
...
-			senderPublicKey: publicKey,
-			moduleAssetId: reclaimTxModuleAssetId,
-		});
-		if (reclaimTx) {
+		let accountInfo;
+		let cachedAccountInfoStr;
+
+		const legacyHexAddress = getLegacyHexAddressFromPublicKey(publicKey);
+
+		// Fetch legacy account info from core, on timeout try querying the cache
```
nagdahimanshu, line 296: Isn't this logic should be opposite? (I mean, to avoid unnecessary calls to `lisk-core`, we can first try to fetch legacy infiormation from cache, if not exists that only call core)

sameersubudhi: The idea is to fetch the information from the Core and use the cache as a fallback. If we use the cache first, we'd never hit the Lisk Core.

MichalTuleja: Definitely cache first, retrieval only on cache miss.

MichalTuleja (changes requested):

services/core/shared/core/compat/sdk_v5/accounts.js
sameersubudhi: @MichalTuleja Sure, I'll then revert the change and then place the cache to avoid making calls to the Core.

nagdahimanshu (approved):
LGTM 👍

MichalTuleja (approved):
👍

PR #679, opened 2021-08-06
Fix produced blocks and rewards calculation - Closes #676

### LINKED ISSUE #676: Fix produced blocks and rewards calculation
### Actual behavior

The account properties `producedBlocks` and `rewards` are incorrect.

### Expected behavior

Make sure the account properties `producedBlocks` and `rewards` are calculated properly.

### Steps to reproduce

To verify, we can compare:

- https://testnet-service-staging.lisk.io/api/v2/blocks?generatorUsername=przemer
- https://testnet-service-staging.lisk.io/api/v2/accounts?username=przemer 

The total block count should match `producedBlocks`. 

`0.4.2`

nagdahimanshu:
### What was the problem?
This PR resolves #676 

### How was it solved?
- [x] Fix `indexAccountsbyPublicKey` method
- [x] Add logic to update delegate cache with `producedBlocks` and `rewards` on every newBlock/deleteBlock event
- [x] Fix transactions `meta.total` when includePending is true

### How was it tested?
- [x] Locally against `testnet`

### DIFF
--- a/services/core/shared/core/compat/sdk_v5/accounts.js
+++ b/services/core/shared/core/compat/sdk_v5/accounts.js
@@ -228,8 +228,8 @@ const resolveDelegateInfo = async accounts => {
 const indexAccountsbyPublicKey = async (accountInfoArray) => {
 	const accountsDB = await getAccountsIndex();
 	const finalAccountsToIndex = await BluebirdPromise.map(
-		dropDuplicates(accountInfoArray
-			.map(accountInfo => getHexAddressFromPublicKey(accountInfo.publicKey))),
+		accountInfoArray
+			.map(accountInfo => getHexAddressFromPublicKey(accountInfo.publicKey)),
 		async address => {
 			const { data: [account] } = await getAccountsFromCore({ address });
 			const [accountInfo] = accountInfoArray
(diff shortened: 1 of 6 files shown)


sameersubudhi:
Kindly consider the following changes:
1. Checking for missing blocks within the `else` section of `checkIndexReadiness` in `services/core/shared/core/compat/sdk_v5/blocks.js`.
```
			} else {
				logger.debug('Blocks index is not yet ready');
				await indexMissingBlocks(getIndexStartHeight(), currentChainHeight);
			}
```
2. Update the delegate cache with the `producedBlocks` and `rewards`information on the `newBlock` and `deleteBlock` events.

sameersubudhi (changes requested):
Please check the above recommendations.

sameersubudhi (changes requested):
We still have an unhandled scenario described below:

An account isn't a delegate anymore when the block containing their delegate registration transaction gets rolled back (`deleteBlock` event). This change should be reflected within our delegate cache as well.

In case of any doubts, we can clarify it offline.

sameersubudhi (approved):

MichalTuleja (approved):

PR #730, opened 2021-08-30
Punished delegates still with same rank - Closes #729

### LINKED ISSUE #729: Punished delegates still with same rank
At the moment, if a delegate is rank 30 and gets punished, it remains at rank 30 but a punished delegate should not ranked with its vote weight. Punished delegates should not be able to keep their prominent rank in top 101. Instead, the vote weight should be changed to 0 LSK as long as the delegate is punished. What ever solution you might find, but a punished delegate should be on the lowest rank.

sameersubudhi:
### What was the problem?
This PR resolves #729 

### How was it solved?
- [x] Set `voteWeight` for both `punished` and `banned` delegates to `0`

### How was it tested?
Local


services/core/shared/core/compat/sdk_v5/accounts.js
```diff
@@ -196,6 +197,15 @@ const resolveAccountsInfo = async accounts => {
 	return accounts;
 };
 
+const verifyIfPunished = async delegate => {
+	const lastestBlockString = await latestBlockCache.get('latestBlock');
```
nagdahimanshu, line 201:
```suggestion
	const latestBlockString = await latestBlockCache.get('latestBlock');
```

services/core/shared/core/compat/sdk_v5/accounts.js
```diff
@@ -196,6 +197,15 @@ const resolveAccountsInfo = async accounts => {
...
+	const lastestBlock = JSON.parse(lastestBlockString);
```
nagdahimanshu, line 202: I think, adding a null check here would be good.
```
const latestBlock = if(latestBlockString) JSON.parse(latestBlockString)

nagdahimanshu (changes requested):

nagdahimanshu (approved):
Looks good 👍

MichalTuleja (approved):

PR #734, opened 2021-08-31
Increase max limit for accounts endpoint to 103 - Closes #733

### LINKED ISSUE #733: Change delegate request limit from 100 to 101
Regarding `/api/v2/accounts?isDelegate=true&limit=100` and https://github.com/LiskHQ/lisk-service/blob/development/docs/api/version2.md#accounts

As the numbers of active delegates is 101 and explorers want to show all 101 in one go, you should increase the limit from 100 to 101.

sameersubudhi:
### What was the problem?
This PR resolves #733 

### How was it solved?
- [x] Increase max `limit` for accounts endpoint in `gateway` to `103`

### How was it tested?
Local


nagdahimanshu (approved):
Looks good 👍

MichalTuleja (approved):

PR #740, opened 2021-09-01
Fix unlocking periods calculation - Closes #735

### LINKED ISSUE #735: Fix unlocking periods calculation
### Actual behavior

Unlocking period is calculated incorrectly for a punished delegate.

`"height": { "start": 14146762, "end": 14406762 }`

### Expected behavior

Unlocking period should be calculated for:
- waiting period
- punishment period

Check this code for a reference:

https://github.com/LiskHQ/lisk-sdk/blob/785c0931e294e36a6abeb3f55640427a980f7957/framework/src/modules/dpos/transaction_assets/unlock_transaction_asset.ts

### Steps to reproduce

Retrieve this account info on testnet: `lskfsk56b28e5byfxmw8944oc3er6bsmq3jag3nw4`

sameersubudhi:
### What was the problem?
This PR resolves #735 

### How was it solved?
- [x] Re-calculate unlocking heights when the delegate is punished

### How was it tested?
Local against the Testnet for account: `lskfsk56b28e5byfxmw8944oc3er6bsmq3jag3nw4`

### DIFF
--- a/services/core/shared/core/compat/sdk_v5/accounts.js
+++ b/services/core/shared/core/compat/sdk_v5/accounts.js
@@ -226,8 +226,22 @@ const resolveDelegateInfo = async accounts => {
 				account.username = account.dpos.delegate.username;
 				account.balance = account.token.balance;
 				account.pomHeights = account.dpos.delegate.pomHeights
-					.sort((a, b) => b - a).slice(0, 5)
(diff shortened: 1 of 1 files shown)


nagdahimanshu (approved):
Looks good and works well 👍

MichalTuleja (approved):

PR #779, opened 2021-09-22
Create regression tests for testnet - Closes #652

### LINKED ISSUE #652: Create smoke tests for testnet
### Description

Create smoke tests in the repository that are able to test basic functionality of the application.

### Motivation

- Smoke tests are able to detect regressions in future deployments

sameersubudhi:
### What was the problem?
This PR resolves #652 

### How was it solved?
- [x] Capture results from `https://testnet-service.lisk.io` and `wss://testnet-service.lisk.io`
- [x] Write jest based regression test cases to compare against the pre-captured results (above) for both:
  - [x] `http`
  - [x] `rpc`

### How was it tested?
Local


tests/regression/api_v2/testnet/http/fees.test.js
```diff
@@ -0,0 +1,43 @@
...
+const baseUrlV2 = `${baseUrl}/api/v2`;
+const endpoint = `${baseUrlV2}/fees`;
+
+describe('Fees API', () => {
+	it('Retrieve fees', async () => {
+		const response = await api.get(`${endpoint}`);
+
+		// response.data
+		// feeEstimatePerByte might vary based on network traffic
+		expect(response.data.feeEstimatePerByte).toStrictEqual(fees.data.feeEstimatePerByte);
```
MichalTuleja, line 33: Why is `.toStrictEqual()` here when the results are non-deterministic?
Can you please consider safe ranges?

tests/regression/api_v2/testnet/rpc/network.test.js
```diff
@@ -0,0 +1,66 @@
...
+		expect(response.result.data.minRemainingBalance)
+			.toStrictEqual(networkStatus.result.data.minRemainingBalance);
+		expect(response.result.data.maxPayloadLength)
+			.toStrictEqual(networkStatus.result.data.maxPayloadLength);
+
+		// response.result.meta
+		expect(response.result.meta.lastUpdate)
+			.toBeGreaterThanOrEqual(networkStatus.result.meta.lastUpdate);
+		expect(Number(response.result.meta.lastBlockHeight))
+			.toBeGreaterThanOrEqual(Number(networkStatus.result.meta.lastBlockHeight));
```
MichalTuleja, line 64: I see some inconsistency here. These `expect()` lines are testing the response and they won't find any extra properties in the future releases. However the `toStrictEqual()` rules in the files above will fail on the new property.

I think the desired behaviour would be to tolerate the extra params, but being strict about the current ones.

MichalTuleja (changes requested):
Please consider the notes added below. The tests written in this PR need a proper discussion for clarification.

PR #852, opened 2021-11-03
Rename delegate status to ineligible - Closes #843

### LINKED ISSUE #843: Delegate status must be renamed to ineligible
### Actual behavior
Delegates whose weight is less than 100 LSK are marked as non-eligible.

### Expected behavior
They should be marked as ineligible.

nagdahimanshu:
### What was the problem?
This PR resolves #843 

### How was it solved?
- [x] Rename delegate status `non-eligible` to `ineligible`
- [x] Update intergation test case and schema
- [x] Update docs

### How was it tested?
Local +Jenkins

### DIFF
--- a/services/core/shared/core/delegates.js
+++ b/services/core/shared/core/delegates.js
@@ -34,7 +34,7 @@ const delegateStatus = {
 	STANDBY: 'standby',
 	BANNED: 'banned',
 	PUNISHED: 'punished',
-	NON_ELIGIBLE: 'ineligible',
+	INELIGIBLE: 'ineligible',
 };
 
 let rawNextForgers = [];
@@ -83,7 +83,7 @@ const computeDelegateStatus = async () => {
 			logger.debug('Determine delegate status');
 
 			// Default delegate status
-			delegate.status = delegateStatus.NON_ELIGIBLE;
+			delegate.status = delegateStatus.INELIGIBLE;
 
 			// Update delegate status, if applicable
 			if (delegate.dpos.delegate.isBanned) {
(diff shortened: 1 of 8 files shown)


services/core/shared/core/delegates.js
```diff
@@ -34,7 +34,7 @@ const delegateStatus = {
 	STANDBY: 'standby',
 	BANNED: 'banned',
 	PUNISHED: 'punished',
-	NON_ELIGIBLE: 'non-eligible',
+	NON_ELIGIBLE: 'ineligible',
```
sameersubudhi, line 37: Please also rename the constant identifier at the necessary places.
```suggestion
	INELIGIBLE: 'ineligible',
```

sameersubudhi (approved):
LGTM 👍

services/gateway/apis/http-version1/methods/delegates.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
 		status: {
 			optional: true,
 			type: 'string',
-			pattern: /^(?:\b(?:active|standby|banned|punished|non-eligible)\b|\b(?:active|standby|banned|punished|non-eligible|,){3,}\b){1}$/,
+			pattern: /^(?:\b(?:active|standby|banned|punished|ineligible)\b|\b(?:active|standby|banned|punished|ineligible|,){3,}\b){1}$/,
```
sameersubudhi, line 33: This change probably needs to be reverted. @MichalTuleja please confirm.

MichalTuleja: It might run into same compatibility issues as the APIv2 change. There's no need to revert it, however it is important to test the APIv2 with clients.

MichalTuleja (commented):
Looks good to me, with the remark written below. It will be approved upon client tests.

MichalTuleja:
This change will not be applied. It would break the API compatibility (version 2).

PR #918, opened 2021-12-14
CORS issue: the headers are skipped when origin is not set - Closes #910

### LINKED ISSUE #910: CORS issue: the headers are skipped when origin is not set
### Actual behavior

The current CORS implementation has an issue: when `origin` is not set the CORS headers are skipped.

### Expected behavior

CORS response is set independently from the `origin` headers.

### Steps to reproduce

1. Call the endpoints with no origin headers.
2. Observe the response.

All

### Additional information

Consider these two requests:

```
francois@sleeper-service ~ $ curl --verbose https://testnet-service.lisk.com/api/v2/network/status 2>&1 | grep -i access-control-


francois@sleeper-service ~ $ curl --verbose --header Origin:http://localhost:1234 https://testnet-service.lisk.com/api/v2/network/status 2>&1 | grep -i access-control-
< access-control-allow-origin: *
< access-control-expose-headers:
< access-control-allow-headers: Content-Type, Access-Control-Request-Method, Access-Control-Request-Headers, Access-Control-Max-Age
< access-control-allow-methods: GET, OPTIONS, POST, PUT, DELETE
< access-control-max-age: 3600
```

sameersubudhi:
### What was the problem?
This PR resolves #910 

### How was it solved?
- [x] Manually set request origin to `http + request_host` if not exists inside moleculer-web's `httpHandler`

### How was it tested?
Local

```
Sameers-MBP:lisk-service sameer$ curl --verbose http://localhost:9902/api/v2/network/status 2>&1 | grep -i access-control-
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: 
< Access-Control-Allow-Headers: Content-Type, Access-Control-Request-Method, Access-Control-Request-Headers, Access-Control-Max-Age
< Access-Control-Allow-Methods: GET, OPTIONS, POST, PUT, DELETE
< Access-Control-Max-Age: 3600

Sameers-MBP:lisk-service sameer$ curl --verbose --header Origin:http://localhost:1234 http://localhost:9902/api/v2/network/status 2>&1 | grep -i access-control
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: 
< Access-Control-Allow-Headers: Content-Type, Access-Control-Request-Method, Access-Control-Request-Headers, Access-Control-Max-Age
< Access-Control-Allow-Methods: GET, OPTIONS, POST, PUT, DELETE
< Access-Control-Max-Age: 3600
```

### DIFF
--- a/services/gateway/shared/moleculer-web/methods.js
+++ b/services/gateway/shared/moleculer-web/methods.js
@@ -23,6 +23,12 @@ module.exports = {
 
 			let requestID = req.headers['x-request-id'];
 			if (req.headers['x-correlation-id']) requestID = req.headers['x-correlation-id'];
+
+			// Manually set request origin to `http + host` if not exists
+			// to ensure proper `Access-Control-*` response headers
+			// TODO: Check for security implications
+			if (!req.headers.origin) req.headers.origin = `http://${req.headers.host}`;
+
 			const options = { requestID };
 			if (this.settings.rootCallOptions) {
 				if (_.isPlainObject(this.settings.rootCallOptions)) {


MichalTuleja (commented):
Based on the specs and the SO thread linked below, the CORS headers are required only when the origin is set. Therefore, there is no need to manually set `origin` for each request that is not a cross-domain request.

To explain it more, please consider the following scenarios:

- The client runs in the browser in the domain `example.com` and performs a request to `service.lisk.com` -> a cross-origin request is being performed -> the `origin` header needs to be set to `example.com` -> the response contains CORS headers
- The request is performed from a command line, HTTP client directly requests data with no `origin` header in the request -> no need to add CORS as the client makes a direct request -> no CORS headers in the response as there is no cross-origin request in this case

https://stackoverflow.com/questions/15988323/cors-and-origin-header
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin

tl;dr As long as the request is performed with `origin` header and CORS headers are set in the response we can consider it a proper behaviour. Direct requests do not require CORS at all.

fchavant:
- I don't think a stackoverflow question/response should be considered a reliable source.
- Since service only provides an API that is always called from another domain I think it'd be fine to always set CORS headers (maybe I'm missing something here? any drawback to doing that)

services/gateway/shared/moleculer-web/methods.js
```diff
@@ -23,6 +23,12 @@ module.exports = {
 
 			let requestID = req.headers['x-request-id'];
 			if (req.headers['x-correlation-id']) requestID = req.headers['x-correlation-id'];
+
+			// Manually set request origin to `http + host` if not exists
+			// to ensure proper `Access-Control-*` response headers
+			// TODO: Check for security implications
+			if (!req.headers.origin) req.headers.origin = `http://${req.headers.host}`;
```
fchavant, line 30: That seems like a bit of a kludge.

MichalTuleja:
@fchavant 

- The second link is a reliable source.
- How would you justify adding CORS headers to the same-origin response? The current behaviour is desired as the same-origin requests do not require `Origin` header and the cross-origin request always have the `Origin` header added by the client. 

Hence, the server response is correct in both cases.

fchavant:
> * How would you justify adding CORS headers to the same-origin response? The current behaviour is desired as the same-origin requests do not require `Origin` header and the cross-origin request always have the `Origin` header added by the client.

I don't think we have any non cross-origin calls. Also I do not think having the CORS headers set all the time is an issue (at least I can't think of an edge case).

PR #924, opened 2022-01-04
Store blockchain search index metadata in MySQL - Closes #913

### LINKED ISSUE #913: Store blockchain search index metadata in MySQL
### Description

Move the index state information from Redis to MySQL.

Once the account import is done, there is no need for re-indexing.
Keep the flag in MySQL database that the genesis account import is done.

### Acceptance Criteria

Create a key-value table in MySQL (core) database.
Ensure the information about the account index completeness is stored there, but only when it reaches 100%.

sameersubudhi:
### What was the problem?
This PR resolves #913 

### How was it solved?
- [x] Add MySQL-based KV-store abstraction
- [x] Add functional tests
- [x] Remove Redis dependency for genesis accounts indexing
- [x] Track genesis accounts indexing progress information on the MySQL-based KV-store
- [x] Ensure genesis account indexing resumes when the API client re-instantiates while the process is in progress
- [x] Query the above KV-store and skip genesis account indexing if the required flag is already set

### How was it tested?
Local


services/core/shared/core/compat/sdk_v5/blockchainIndex.js
```diff
@@ -320,18 +336,27 @@ const performGenesisAccountsIndexing = async () => {
...
 		return;
 	}
 
 	try {
 		logger.info('Attempting to update genesis account index (one-time operation)');
 		await performGenesisAccountsIndexing();
-		await genesisAccountsCache.set('isGenesisAccountIndexingScheduled', true);
-		await genesisAccountsCache.set('genesisAccountPageCached', 0);
+		await genesisAccountsCache.set(isGenesisAccountIndexingScheduled, true);
+		await genesisAccountsCache.set(genesisAccountPageCached, 0);
```
nagdahimanshu, line 359: Don't we need to completely move account index information from redis to MySQL)?

sameersubudhi: @MichalTuleja Please confirm.

MichalTuleja: I think that would even work assuming that information reflects the number of imported accounts, not the number of scheduled accounts. That's why I see an issue here: the import is scheduled but not yet in the DB snapshot. For the DB snapshot we need another flag that indicates the account import has already finished. Is the `genesisAccountPageCached` indicating the number of page cached or scheduled?

sameersubudhi: The Redis dependency has now been completely removed from the process.

services/core/shared/core/compat/sdk_v5/blockchainIndex.js
```diff
@@ -285,16 +292,20 @@ const cacheLegacyAccountInfo = async () => {
...
 		.map(account => account.address);
-	const genesisAccountPageCached = 'genesisAccountPageCached';
 
 	logger.info(`${genesisAccountsToIndex.length} registered accounts found in the genesis block`);
 
-	const lastCachedPage = await genesisAccountsCache.get(genesisAccountPageCached) || 0;
+	const [{ value: lastCachedPageStr } = {}] = await keyValueDB.find(
+		{ key: genesisAccountPageCached, limit: 1 },
+		['value'],
+	);
```
MichalTuleja, line 307: These `get/set` functions worked pretty well before. Considering we're switching to MySQL and treat one table as key-value store, can you please add a little bit abstraction here? I mean put that logic into the get/set functions and we can consider it ready.

services/core/shared/core/compat/sdk_v5/blockchainIndex.js
```diff
@@ -320,18 +344,27 @@ const performGenesisAccountsIndexing = async () => {
...
 const indexGenesisAccounts = async () => {
-	const isScheduled = await genesisAccountsCache.get('isGenesisAccountIndexingScheduled');
+	const isGenesisAccountsAlreadyIndexed = async () => {
+		const keyValueDB = await getKeyValueStoreIndex();
+		const [{ value: isAlreadyIndexed } = {}] = await keyValueDB.find(
+			{ key: isGenesisAccountIndexingFinished, limit: 1 },
+			['value'],
+		);
+		return isAlreadyIndexed === String(true); // since 'value' is of type 'string'
+	};
```
MichalTuleja, line 354: Same as above.

services/core/shared/core/compat/sdk_v5/blockchainIndex.js
```diff
@@ -307,11 +318,24 @@ const performGenesisAccountsIndexing = async () => {
...
+
+			// Update MySQL based KV-store to avoid re-indexing of the genesis accounts
+			// after applying the DB snapshots
+			if (pageNum === NUM_PAGES - 1) {
+				logger.info('Setting genesis account indexing completion status in MySQL DB');
+				await keyValueDB.upsert({
+					key: isGenesisAccountIndexingFinished,
+					value: String(true),
+				});
+			}
```
MichalTuleja, line 338: Same as above.

MichalTuleja (changes requested):

nagdahimanshu (approved):
LGTM 👍

MichalTuleja (approved):

PR #932, opened 2022-01-12
Fix votes_received inconsistencies - Closes #921

### LINKED ISSUE #921: 'votes_received' endpoint returns different data across Lisk Services
### Actual behavior

It appears that service.lisk.com and mainnet-service.lisktools.eu return two different arrays of voters. It seems like service.lisk.com is wrong and mainnet-service.lisktools.eu to be correct.

```
"meta": {
    "count": 62,
    "offset": 0,
    "total": 62
  }
https://service.lisk.com/api/v2/votes_received?username=carbonara&aggregate=true&limit=100&offset=0
```
```
"meta": {
    "count": 61,
    "offset": 0,
    "total": 61
  }
https://mainnet-service.lisktools.eu/api/v2/votes_received?username=carbonara&aggregate=true&limit=100&offset=0
```

The extra address seems to be at least `lskm7dquz6jepotwx4z4uyt2oswaaphdyz98nzy7c`. If I'm not wrong, there is at least another address from another delegate affected from this issue.

### Expected behavior

Probably the right data is from mainnet-service.lisktools.eu

### Steps to reproduce
see above.

API - version unknown.

sameersubudhi:
### What was the problem?
This PR resolves #921 

### How was it solved?
- [x] Use `for` loop instead of `Bluebird.map`
- [x] Query `votesDB` and `votesAggregateDB` with `limit` to avoid warning 
- [x] Fix functional test for MySQL

### How was it tested?
Local


MichalTuleja (approved):
Excellent work, thanks @sameersubudhi

PR #974, opened 2022-02-18
Miscalculation issues with total fee - Closes #971

### LINKED ISSUE #971: Miscalculation issues with total fee
### Actual behavior

The total fee value is wrongly calculated for multisignature transactions.

https://service.lisk.com/api/v2/blocks?height=17760266

Lisk Service uses the SDK's API Client that apparently, does not give precise results when some of the public keys are empty.

```
"totalForged": "100217000",
"totalBurnt": "281000",
"totalFee": "-64000",
```

### Expected behavior

The returned values are positive and correct.

### Steps to reproduce

Send a multisig transaction with minimum required number of optional signatures.

https://service.lisk.com/api/v2/blocks?height=17760266

`0.6.0`

### Additional information

One of the possible solutions might be to replace the API Client `apiClient.transaction.computeMinFee(txn)` with the corresponding SDK transaction module.

The change should be done in the line 95 of `blocks.js`.

https://github.com/LiskHQ/lisk-service/blob/4591de5eff40368aca9785d4f31ea2cd292d6fdf/services/core/shared/core/compat/sdk_v5/blocks.js#L95

sameersubudhi:
### What was the problem?
This PR resolves #971 

### How was it solved?
- [x] Add in-memory caching for transactions schemas
- [x] Add in-memory caching for genesis config
- [x] Use `computeMinFee` directly from Lisk elements (`@liskhq/lisk-transactions`) instead from the apiClient
- [x] Add tests

### How was it tested?
Local (`totalFee` for the block at height `17760266` on the mainnet should be non-negative, exactly `0`)


services/core/shared/core/compat/sdk_v5/blocks.js
```diff
@@ -90,17 +88,30 @@ const normalizeBlocks = async (blocks, includeGenesisAccounts = false) => {
...
+						},
+					);
+
+					block.size += txn.size;
+
+					const txnMinFee = BigInt(txn.minFee);
+					block.totalForged += BigInt(txn.fee);
+					block.totalBurnt += txnMinFee;
+					block.totalFee += BigInt(txn.fee) - txnMinFee;
+				},
```
MichalTuleja, line 112: Please wrap this logic into a named function.

MichalTuleja (changes requested):

services/core/shared/core/compat/sdk_v5/blocks.js
sameersubudhi: Since the current code logic also affects the `block` object while acting only on the block payload (transactions), refactored the code to calculate the transaction min fee into a named function instead.

ManuGowda (approved):
@sameersubudhi the code changes LGTM, lets test it manually to ensure fees calculated properly

sameersubudhi:
@ManuGowda Refactored the code and added tests. Had to add them to the functional tests because of the transitive dependency on Redis within `services/core/shared/core/compat/sdk_v5/coreApi.js`.

ManuGowda (changes requested):
@sameersubudhi overall the changes looks good
- please update package lock to have https://registry.npmjs.org/
- the naming of directory for `constants` should be `fixtures`?
rest looks good

ManuGowda (approved):

ManuGowda (approved):

MichalTuleja (approved):

PR #991, opened 2022-03-11
Fix false error issue for range based queries - Closes #990

### LINKED ISSUE #990: False error when providing the correct amount range
### Actual behavior
When querying https://service.lisk.com/api/v2/transactions?amount=11:100&limit=10
service returns `{"error":true,"message":"From amount cannot be greater than to amount."}`
### Expected behavior
Service should return transactions within the given amount range
### Steps to reproduce
Query https://service.lisk.com/api/v2/transactions?amount=11:100&limit=10
Service 0.6.1

nagdahimanshu:
### What was the problem?
This PR resolves #990 

### How was it solved?
- [x] Cast to Number before performing a check
- [x] Add unit test cases

### How was it tested?
Local + Jenkins

### DIFF
--- a/services/core/shared/core/compat/sdk_v5/paramUtils.js
+++ b/services/core/shared/core/compat/sdk_v5/paramUtils.js
@@ -19,7 +19,8 @@ const {
 
 const normalizeRangeParam = (params, property) => {
 	if (typeof params[property] === 'string' && params[property].includes(':')) {
-		const [from, to] = params[property].split(':');
+		const from = Number(params[property].split(':')[0]);
+		const to = Number(params[property].split(':')[1]);
 		if (from && to && from > to) throw new ValidationException(`From ${property} cannot be greater than to ${property}.`);
 
 		if (!params.propBetweens) params.propBetweens = [];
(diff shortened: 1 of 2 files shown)


services/core/shared/core/compat/sdk_v5/paramUtils.js
```diff
@@ -19,7 +19,8 @@ const {
 
 const normalizeRangeParam = (params, property) => {
 	if (typeof params[property] === 'string' && params[property].includes(':')) {
-		const [from, to] = params[property].split(':');
+		const from = Number(params[property].split(':')[0]);
+		const to = Number(params[property].split(':')[1]);
```
sameersubudhi, line 23: Kindly consider including the following check and a corresponding test case for the negative scenario.

```suggestion
		const [fromStr, toStr] = params[property].split(':');
		const from = Number(fromStr);
		const to = Number(toStr);
		if (Number.isNaN(from) || Number.isNaN(to)) throw new ValidationException(`Invalid (non-numeric) '${property}' range values supplied: ${params[property]}.`);
```

sameersubudhi (changes requested):

PR #1032, opened 2022-03-28
Create Blockchain Connector - Closes #992

### LINKED ISSUE #992: Create Blockchain Connector
### Description

Clone the `core-service` and keep the lowest SDK-related abstraction layer.

Make the service methods exposed by the framework methods programmatically.

Avoid manual mappings where unnecessary.

Migrate the GeoIP service bindings.

Lisk Service-Planned arch drawio (8)

### Acceptance Criteria

- Blockchain connector can retrieve data from the SDK application
- Methods mappings reflect the SDK application
- Methods exports are done to other microservices
- The startup sequence is preserved

### Additional information

Make sure the start-up sequence works as follows:

1. Connect to the blockchain application
1. Retrieve enabled modules
1. Retrieve available transaction types
1. Determine the current block height
1. Subscribe to the built-in block-related events (newBlock, etc.)
1. Subscribe to the extra events described in the schema
1. Connect to the GeoIP service (optional, when available)
1. Refresh data based on the events coming from the blockchain-connector

sameersubudhi:
### What was the problem?
This PR resolves #992 

### How was it solved?
- [x] Create `blockchain-connector` microservice. Registered it with the message broker with the name `connector`.
- [x] Update `Makefile` to include the new microservice
- [x] Update framework to `v1.3.1` to:
  - [x] Resolve exported promises
  - [x] Support both targeted/non-targeted emits and broadcasting of events
- [x] Update all microservices to use the new framework
- [x] Add custom methods/events (with auto-decoded payload) for all the actions/events defined by SDK (always available when an application is built with the SDK)
- [x] Dynamically export all the methods and events exposed by the SDK application
- [x] Implement `encoder` and `decoder` to automatically serialise and deserialise the blockchain entities without the need for an active apiClient

### How was it tested?
Local


services/blockchain-connector/src/sdk_v5/helper/decoder.js
```diff
@@ -0,0 +1,125 @@
...
+			...blockHeader,
+			asset: blockHeaderAsset,
+		},
+		payload: blockPayload,
+	};
+	return decodedBlock;
+};
+
+module.exports = {
+	getAccountSchema,
```
nagdahimanshu, line 115: I would recommend moving those schema-related functions to a separate file.

nagdahimanshu, line 115: I would recommend moving those schema-related functions to a separate file.

sameersubudhi: Done!

nagdahimanshu (approved):
Looks good 👍

nagdahimanshu (approved):

MichalTuleja (approved):

PR #1069, opened 2022-05-02
Fix normalizeRangeParam method - Closes #1056

### LINKED ISSUE #1056: Fix normalizeRangeParam method
### Actual behavior

`normalizeRangeParam` method in [paramUtils.js](https://github.com/LiskHQ/lisk-service/pull/1015/files/f825f92dd33b1000023f56383a74f9b22b133e48#diff-1f48143e001fdafa81ecae39c0bf4c92f9ddaeb42aaae0915538a0e12109dd58) currently does not support input params like (`-10000:0`, `100:` or `:100`) appropriately. In its current form, the unsupplied `to` or `from` values default to `0` instead of being open-ended.

### Expected behavior
`normalizeRangeParam` should not default the `to` or `from` values to `0` when the user input does not explicitly specify the values.

### Steps to reproduce
Querying for downvotes with https://service.lisk.com/api/v2/transactions?amount=-1000000:0 returns transactions with upvotes (positive amount values).

### Fix
Update `normalizeRangeParam`:
```
		if (fromStr === '') {
			params.propBetweens.push({ property, to });
		} else if (toStr === '') {
			params.propBetweens.push({ property, from });
		} else {
			params.propBetweens.push({ property, from, to });
		}
```

And, update logic for `params.propBetweens` in `mysql.js` to:
```
					if ('from' in propBetween) query.where(propBetween.property, '>=', propBetween.from);
					if ('to' in propBetween) query.where(propBetween.property, '<=', propBetween.to);
					if ('greaterThan' in propBetween) query.where(propBetween.property, '>', propBetween.greaterThan);
					if ('lowerThan' in propBetween) query.where(propBetween.property, '<', propBetween.lowerThan);
```

`v0.6.2`

sameersubudhi:
### What was the problem?
This PR resolves #1056 

### How was it solved?

- [x] Update `normalizeRangeParam` method to include `from` and `to` values in `propBetween` only when defined
- [x] Update `mysql.js` to support the above change in `core` and `newsfeed` microservices
- [x] Add unit tests for the introduced changes

### How was it tested?
Local

### DIFF
--- a/services/core/shared/core/compat/sdk_v5/paramUtils.js
+++ b/services/core/shared/core/compat/sdk_v5/paramUtils.js
@@ -17,6 +17,8 @@ const {
 	Exceptions: { ValidationException },
 } = require('lisk-service-framework');
 
+const EMPTY_STRING = '';
+
 const normalizeRangeParam = (params, property) => {
 	if (typeof params[property] === 'string' && params[property].includes(':')) {
 		const [fromStr, toStr] = params[property].split(':');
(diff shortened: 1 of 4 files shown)


nagdahimanshu (approved):
Looks good to me 👍

PR #1075, opened 2022-05-16
Inconsistent aggregated values - Closes #1072

### LINKED ISSUE #1072: Inconsistent aggregated values
### Actual behavior
The aggregated values for the `/accounts` (producedBlocks/rewards) and `/votes_received` (voters/aggregated votes) endpoints are inconsistent.

### Expected behavior
All the aggregated values should be consistent.

### Steps to reproduce
Run Lisk Service against the mainnet and wait for it to finish indexing.
Verify the aggregated values against the blocks (for producedBlocks/rewards) and transactions (for voters/aggregated votes) index for the particular addresses.

`release/0.6.3`

sameersubudhi:
### What was the problem?
This PR resolves #1072 

### How was it solved?
- [x] Update `updateAccountsCacheListener` to use queues instead of directly updating the index
- [x] Ensure `getGeneratorPkInfoArray` always receives `generatorPublickey` within input params
- [x] Update `knex` version and switch MySQL client from `mysql` to `mysql2`
- [x] Wait to start indexing if the Lisk Core node is still syncing with the network

### How was it tested?
Index mainnet on multiple instances
 - Compare `producedBlocks` from `/accounts` endpoint against `meta.total` from `/blocks` endpoint for any given account
 - Compare `/accounts` response on the two deployments
 - Compare `/votes_received` response on the two deployments

### DIFF
--- a/services/core/shared/core/compat/sdk_v5/blockchainIndex.js
+++ b/services/core/shared/core/compat/sdk_v5/blockchainIndex.js
@@ -189,18 +189,19 @@ const ensureArray = (e) => Array.isArray(e) ? e : [e];
 const indexBlock = async job => {
 	const { height } = job.data;
 
-	const blocksDB = await getBlocksIndex();
 	const blocks = await getBlockByHeight(height);
+	if (!validateBlocks(blocks)) throw new Error(`Error: Invalid block at height ${height}`);
+
+	const blocksDB = await getBlocksIndex();
+	const transactionsDB = await getTransactionsIndex();
+	const votesDB = await getVotesIndex();
+	const multisignatureDB = await getMultisignatureIndex();
+
 	const connection = await getDbConnection();
 	const trx = await startDbTransaction(connection);
+	logger.trace(`Created new MySQL transaction to index block at height ${height}`);
 
-	logger.debug(`Created new MySQL transaction to index block at height ${height}`);
-
-	if (!validateBlocks(blocks)) throw new Error(`Error: Invalid block at height ${height}`);
 	try {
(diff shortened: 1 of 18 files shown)


services/core/shared/core/compat/sdk_v5/blockchainIndex.js
```diff
@@ -244,14 +250,17 @@ const updateBlockIndex = async job => {
 };
 
 const deleteIndexedBlocks = async job => {
```
nagdahimanshu, line 252: As the given method only delete a single block, in my opinion, it would be better to rename this method to `deleteIndexedBlock` and update the places where it has been called.

sameersubudhi: No, this method is capable of deleting multiple blocks from the index and is used for the same within the `verifyAndIndexBlock` method a fork is detected. Please check line number 336:

```
		// Fork handling
		if (blockInfo && blockInfo.id !== block.id) {
			const [highestIndexedBlock] = await blocksDB.find({ sort: 'height:desc', limit: 1 }, ['height']);
			const blocksToRemove = await blocksDB.find(
				{
					propBetweens: [{
						property: 'height',
						from: block.height + 1,
						to: highestIndexedBlock.height,
					}],
					sort: 'height:desc',
					limit: highestIndexedBlock.height - block.height,
				},
				['id', 'generatorPublicKey'],
			);
			await deleteIndexedBlocksQueue.add({ blocks: blocksToRemove });
		}
```

nagdahimanshu (approved):
Looks good 👍

PR #1174, opened 2022-07-01
 Fix legacy cache handling - Closes #1170

### LINKED ISSUE #1170: Fix legacy cache handling
### Actual behavior
The legacyAccountCache is not cleared when indexing the chain for legacy addresses that already have reclaimed the balance.
This leads to confusion where accounts that are victims of address collision attacks are led to believe that they are still eligible to reclaim their previous balance but when posting the transactions to the network it fails.

### Expected behavior
The legacyAccountCache should be cleared when the previous blocks are being indexed taking into account all the reclaim transactions.

`v0.6.3`

sameersubudhi:
### What was the problem?
This PR resolves #1170 

### How was it solved?
- [x] Ensure legacy account cache is updated every new block

### How was it tested?
Against the mainnet

### DIFF
--- a/services/core/shared/core/compat/sdk_v5/accounts.js
+++ b/services/core/shared/core/compat/sdk_v5/accounts.js
@@ -639,6 +639,7 @@ const removeReclaimedLegacyAccountInfoFromCache = () => {
 			});
 		}
 	};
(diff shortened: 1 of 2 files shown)


nagdahimanshu (approved):
Looks good 👍

PR #1252, opened 2022-09-01
Update fee estimator - Closes #1227

### LINKED ISSUE #1227: Update fee estimator
### Description

The algorithm now does not require considering the `baseFees` for calculating the fee estimation per byte as per the latest changes in https://github.com/LiskHQ/lips/pull/156.

### Acceptance Criteria

- Remove any `baseFees` (`nameFees`) related logic in the dynamic fee estimator algorithm
- The algorithm should now only consider `minFee` for a transaction
- Update the integration/unit tests, if applicable

priojeetpriyom:
### What was the problem?
This PR resolves #1227

### How was it solved?
- [x] Remove baseFee param from computeMinFee call into lisk-transactions
- [x] Update dependencies
- [x] Fix small bugs on fee estimation
- [x] Check and update test if necessary
### How was it tested?
- [x] Local
 1. Start Lisk Core
 2. update ecosystem.config.js to run - connector, coordinator, fee-estimator, gateway
 3. Run lisk service
 4. Query following api for response: http://localhost:9901/api/v3/fees


services/fee-estimator/shared/dynamicFeesUtils.js
```diff
@@ -129,6 +129,7 @@ const getEstimateFeeByte = async () => { // aka getBestEstimateAvailable
 		status: 'SERVICE_UNAVAILABLE',
 	};
 };
+
 Signals.get('newBlock').add(async () => {
```
sameersubudhi, line 133: Could you please refactor the anonymous listener function (from within the `Signals.get('newBlock').add(...)`) to be a named function and then register it as a named-listener, just like everywhere else?
We do not want any anonymous listeners registered to the signals.

priojeetpriyom: Applied

services/fee-estimator/shared/dynamicFees.js
```diff
@@ -34,14 +34,22 @@ const calculateBlockSize = async block => {
...
+	const wSumLastBlocks = blockSizes.reduce(
+		(prevWSum = 0, blockSize) => {
+			const currentWSum = prevWSum + (blockSize * weight);
+			weight *= (1 - decayFactor);
+			return currentWSum;
+		},
+	);
 
+	const wavgLastBlocks = wSumLastBlocks / blocks.length;
 	return wavgLastBlocks;
```
sameersubudhi, line 56: Please get the change confirmed with someone from the research team as well.
Also, please get your doubt clarified about the following logic in the `getEstimateFeeByteForBlock` method.
```
if (wavgBlockBatch > (12.5 * 2 ** 10) || sizeLastBlock > (14.8 * 2 ** 10)) {
```

sameersubudhi: This comment is invalid!

Updating `const wavgLastBlocks = wSumLastBlocks / blocks.length;` to `const wavgLastBlocks = wSumLastBlocks / totalWeight;` clarifies it.

sameersubudhi (approved):
Great job on your first PR @priojeetpriyom!! Keep it up 🚀

services/fee-estimator/shared/dynamicFees.js
```diff
@@ -34,14 +34,25 @@ const calculateBlockSize = async block => {
...
+	);
+	const decayFactor = 1 - (config.feeEstimates.wavgDecayPercentage / 100);
 	let weight = 1;
-	const wavgLastBlocks = blockSizes.reduce((a = 0, b = 0) => {
-		weight *= 1 - decayFactor;
-		return a + (b * weight);
-	});
+	let totalWeight = 0;
+
+	const wSumLastBlocks = blockSizes.reduce(
```
ricott1, line 46: Personal taste: I would use no abbreviation in variables like this and rather write `weightedSum` just o clarify the code as much as possible

ricott1: In the same spirit, I find this reduce function a bit confusing: `wBlockAcc` is returned, while `weight` is updated in place. Anyway, it looks okay to me after reading the reduce docs, but it would be nice to try the following case:   array of 20 blocks all with the same size. In this case the avg should equal the size of the blocks.

ricott1 (approved):

services/fee-estimator/shared/dynamicFees.js
sameersubudhi: @priojeetpriyom Could you please address the above points in the PR?

priojeetpriyom: The pr is addressed. Thanks for the test suggestion @ricott1 🥇

nagdahimanshu (approved):
Looks good 👍

PR #1270, opened 2022-09-23
Improve token metadata indexing and resolution - Closes #1255

### LINKED ISSUE #1255: Improve token metadata indexing and resolution
### Description

Currently, the `blockchain-app-registry` directly indexes the `tokenID` for the token metadata as specified in the `nativetokens.json`. Instead, the index should persist the tokens by their `localID` along with the `chainID` so that when a user queries for token information either by their `global tokenID` or `local tokenID along with chainID` the API should successfully return the metadata information for the token.

### Acceptance Criteria

- Rename the tables:
  - `applications` --> `application_metadata`
  - `tokens` --> `token_metadata`
- Index `localID`(the least significant 4 bytes of the `tokenID`) instead of `tokenID` and in lower-case (hex characters)
- The following request query param validations are in place:
  - Request is directly processed when the query params only contain the `global tokenID`
  - A validation error is returned when the `global tokenID` doesn't begin with the supplied `chainID`
  - Both `chainID` and `tokenID` are necessary when the supplied `tokenID` value is of type local (the most significant 4 bytes are all zeros)
- Integration tests are added for the above scenarios

priojeetpriyom:
### What was the problem?
This PR resolves #1255

### How was it solved?
- [x] Rename the tables:
  - [x] `applications` --> `application_metadata`
  - [x] `tokens` --> `token_metadata`
- [x] Index `localID`(the least significant 4 bytes of the `tokenID`) instead of `tokenID` and in lower-case (hex characters)
- [x] The following request query param validations are in place:
  - [x] Request is directly processed when the query params only contain the `global tokenID`
  - [x] A validation error is returned when the `global tokenID` doesn't begin with the supplied `chainID`
  - [x] Both `chainID` and `tokenID` are necessary when the supplied `tokenID` value is of type local (the most significant 4 bytes are all zeros)
- [x] Integration tests are added for the above scenarios

### How was it tested?
- [x] Local


sameersubudhi (approved):
Thank you for your contribution @priojeetpriyom! LGTM 👍

Issue #1188, opened 2022-07-20
Improve MySQL implementation


sameersubudhi:
### Description

The current implementation for the `increment` and `decrement` methods within `framework/src/mysql.js` is very basic. They can be improved upon by already utilizing the `queryBuilder` method.

### Acceptance Criteria

- Integrate the `queryBuilder` method in `framework/src/mysql.js` into the `increment` and `decrement` methods
- Update existing/add new functional test cases to test the updated implementation
- Update the `getTableInstance` method in `mysql.js` to read `tableName` directly from the schema file
- Update the `count` and `find` methods to handle composite primary keys
- Update the `count` method to support custom (user-supplied) column-based count
  - Update implementation in `services/blockchain-app-registry/shared/metadata.js` (L91, L192, L313)
- Refactor `count` method to act as a wrapper for `find` method and remove duplicated logic 
- Support `DISTINCT` in the `count` and the `find` methods
- Move the MySQL-based kvStore implementation to the lisk-service-framework
- Rename all methods to use `DB` instead of `Db`
- Add unit test cases for whereJsonSupersetOf
- Add transactional support on `find` and `KVStore.get`
- Move KVStore implementation to framework
- Add support for `IS NULL, IS NOT NULL, LIKE` to the query builder

PR #1279, opened 2022-10-07
Update events indexing - Closes #1262

### LINKED ISSUE #1262: Update events indexing
### Description

With Lisk SDK v6.0.0-alpha.3 and above, the blockchain now emits events and to serve user requests with accurate information for various endpoints, it is important that the information emitted in the payload is indexed properly.

### Acceptance Criteria

The following events are indexed properly for the following activities/entities:
- Multisig registration
- Reward
  - Address the TODO corresponding the `block.reward`
- Validators
- Set ‘executionStatus’ for transactions appropriately

sameersubudhi:
### What was the problem?
This PR resolves #1262 

### How was it solved?
- [x] Enable events indexing
- [x] Remove mocking for `/api/v3/events` endpoint
- [x] Cache `nodeInfo` in `blockchain-connector` and update upon state change
- [x] Fix and plug the following auto-decoders in `blockchain-connector`
  - [x] `decodeTransaction`
    - [x] Now includes `size` and `minFee`
  - [x] `decodeBlock`
  - [x] `decodeEvent`
- [x] Track `totalRewards` and `producedBlocks` for block generators
- [x] Fix `forgedBlocks` mapping for delegates response
- [x] Determine and index `executionStatus` for transactions
- [x] Extend transactions response to include `size` and `minFee`
  - [x] Update swagger
- [x] Code cleanup / refactor

### How was it tested?
- Locally against Lisk Core v4.0.0-alpha.2


services/blockchain-connector/shared/sdk/events.js
```diff
@@ -53,36 +53,15 @@ const subscribeToAllRegisteredEvents = async () => {
...
-		);
-
-		return eventsResponse;
-	} catch (err) {
-		if (err.message.includes(timeoutMessage)) {
-			throw new TimeoutException('Request timed out when calling \'chain_getEvents\'');
-		}
-		throw err;
-	}
+const getEventsByHeightDecoded = async (height) => {
```
priojeetpriyom, line 56: An alternative name could be `getDecodedEventsByHeight`

services/blockchain-connector/shared/sdk/index.js
```diff
@@ -13,12 +13,111 @@
...
+
+	// Cache all the schemas
+	setSchemas(await getSchemas());
+	setMetadata(await getSystemMetadata());
+
+	// Download the genesis block, if applicable
+	await getGenesisBlock();
+
+	// Register listener to update the nodeInfo cache only when it updates
+	const updateNodeInfoCacheListener = getNodeInfo.bind(null, true);
```
priojeetpriyom, line 77: Just curious if it could be written like below as well?

```suggestion
	const updateNodeInfoCacheListener = () => getNodeInfo(true);
```

services/blockchain-connector/shared/utils/parser.js
```diff
@@ -38,7 +40,10 @@ const parseInputBySchema = (input, schema) => {
 	if (typeof input !== 'object') {
 		if (schemaDataType === 'string') return String(input);
 		if (schemaDataType === 'boolean') return Boolean(input);
-		if (schemaDataType === 'bytes') return Buffer.from(input, 'hex');
+		if (schemaDataType === 'bytes') {
+			if (schema.format === 'lisk32') { return getAddressFromLisk32Address(input); }
+			return Buffer.from(input, 'hex');
```
priojeetpriyom, line 45: Curious, why do we convert hex to buffer here?

services/blockchain-indexer/shared/database/schema/eventTopics.js
```diff
@@ -15,8 +15,9 @@
  */
 module.exports = {
 	tableName: 'event_topics',
-	primaryKey: ['id', 'topic'],
+	primaryKey: 'tempID',
 	schema: {
+		tempID: { type: 'string' },
```
priojeetpriyom, line 20: Would naming it `id_topic` be more expressive? Also, why do we need to store it separately since both `id` and `topic` are stored already?

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -118,10 +128,21 @@ const INDEX_VERIFIED_HEIGHT = 'indexVerifiedHeight';
 
 const validateBlock = (block) => !!block && block.height >= 0;
 
+const getTransactionExecutionStatus = (tx, events) => {
+	// TODO: Update implementation with next SDK release and return with constants from SDK
+	// const expectedEventName = `${transaction.module}commandExecutionResult`;
```
priojeetpriyom, line 133: just guessing

```suggestion
	// const expectedEventName = `${tx.module}:commandExecutionResult`;
```

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -118,10 +128,21 @@ const INDEX_VERIFIED_HEIGHT = 'indexVerifiedHeight';
...
+	const expectedEventName = `${tx.module}:transaction`;
+	const commandExecResultEvents = events.filter(e => `${e.module}:${e.name}` === expectedEventName);
+	const txExecResultEvent = commandExecResultEvents.find(e => e.topics.includes(tx.id));
+	return txExecResultEvent.data === '0801' ? 'success' : 'fail';
```
priojeetpriyom, line 137: Maybe we can define a constant for the data `0801`.

services/blockchain-connector/app.js
```diff
@@ -68,12 +64,6 @@ nodeStatus.waitForNode().then(async () => {
 	app.run()
 		.then(async () => {
 			await init();
-
-			const schemas = await getSchemas();
-			setSchemas(schemas);
-
-			// Download the genesis block, if applicable
-			await getGenesisBlock();
```
nagdahimanshu, line 76: Don't we need to download genesis block during start if applicable?

services/blockchain-connector/shared/sdk/events.js
sameersubudhi: I'd prefer to keep the name as is so that when we search for implementations by the original method's name these methods also show up in the result.

services/blockchain-connector/shared/utils/parser.js
sameersubudhi: Because the schema expects the value to be an instance of Buffer and in the serialized form the value is a hex string.

services/blockchain-indexer/shared/database/schema/eventTopics.js
sameersubudhi: This is necessary because of the implementation of our current MySQL abstraction. The `count` method fails if the primary key is composite. This should be fixed with the issue: https://github.com/LiskHQ/lisk-service/issues/1188

services/blockchain-connector/shared/sdk/events.js
priojeetpriyom: This one is already mapped to different name [here](https://github.com/LiskHQ/lisk-service/blob/efa7795cc5039045fc6f4a29ea85f64884ef2035/services/blockchain-connector/shared/sdk/events.js#L66).

services/blockchain-connector/shared/sdk/index.js
priojeetpriyom: yeah, asked for self clarification.

services/blockchain-indexer/shared/indexer/blockchainIndex.js
sameersubudhi: There's a TODO above to address the same. This most likely will be exposed by the SDK as a constant in the next release. I'd leave it as is for now to avoid rework.

services/blockchain-connector/app.js
sameersubudhi: I've moved all the initialization logic within the `init` method in `services/blockchain-connector/shared/sdk/index.js`.

services/blockchain-indexer/shared/database/schema/eventTopics.js
priojeetpriyom: Is it feasible to write raw queries instead of storing redundant data?

nagdahimanshu:
As we removed all encoding/decoding logic to the connector service, we can also remove `lisk-codec` from the indexer as a dependency.

services/blockchain-indexer/shared/database/schema/eventTopics.js
sameersubudhi: I'd like to minimize the changes/rework that would be necessary everywhere in the code once https://github.com/LiskHQ/lisk-service/issues/1188 is addressed. Hence, avoiding the raw queries until and unless that is the only way to go ahead.

priojeetpriyom: I get your point. My only concern is changing DB structure will probably require downtime and will require unnecessary resources until that card is merged. Whereas code changes won't have an additional impact on production downtime.

priojeetpriyom (approved):

services/blockchain-indexer/shared/database/schema/eventTopics.js
sameersubudhi: Thank you for the concern but I'll take care of coordinating and handling the situation when the time comes 🙂

nagdahimanshu (approved):
Looks goos 👍 👏

Issue #1301, opened 2022-10-18
Unable to sort blocks in ascending


sameersubudhi:
### Actual behavior
Querying `/api/v3/blocks` endpoint with `sort=height:asc` or `sort=timestamp:asc` fails.

```
{
  "error": true,
  "message": "Data not found"
}
```

### Expected behavior
Must return properly sorted data.

### Steps to reproduce
- Query `/api/v3/blocks` endpoint with `sort=height:asc` or `sort=timestamp:asc`.

### Which version(s) does this affect? (Environment, OS, etc...)
`v0.7.0-alpha.1`

PR #1308, opened 2022-10-20
Filtering block assets by module fails - Closes #1300 #1301

### LINKED ISSUE #1300: Filtering block assets by module fails
### Actual behavior
Filtering block assets by module fails.

Query: `/api/v3/blocks/assets?module=random`
Response: 
```
{
  "error": true,
  "message": "Server error: select count(`height`) as `count` from `blocks` where `height` is not null and (json_contains(`assetsModules`,'random')) - Invalid JSON text in argument 1 to function json_contains: \"Invalid value.\" at position 0."
}
```

### Expected behavior
The query should return a successful response.

### Steps to reproduce
- Query the endpoint with `module` request param

`v0.7.0-alpha.1`

nagdahimanshu:
### What was the problem?
This PR resolves #1300 #1301 

### How was it solved?
- [x] Fix `getBlocksByIDs` implementation
- [x] Fix regex for module name
- [x] Fix MySql whereJsonSupersetOf implementation
- [x] Publish framework to GitHub
- [x] Update all services dependencies

### How was it tested?
Local


framework/src/mysql.js
```diff
@@ -243,9 +243,9 @@ const getTableInstance = async (tableName, tableConfig, connEndpoint = CONN_ENDP
 			const { property, values } = params.whereJsonSupersetOf;
 			query.where(function () {
 				const [val0, ...remValues] = Array.isArray(values) ? values : [values];
-				this.whereJsonSupersetOf(property, val0);
+				this.whereJsonSupersetOf(property, [val0]);
```
priojeetpriyom, line 246: Can we rename `property` to `column` in respect to knex api?

services/gateway/shared/regex.js
```diff
@@ -34,6 +34,7 @@ const TOKEN_ID = /^\b[a-fA-F0-9]{16}\b$/;
 const BLS_KEY = /^\b[a-fA-F0-9]{96}\b$/;
 const PROOF_OF_POSSESSION = /^\b[a-fA-F0-9]{192}\b$/;
 const TRANSACTION = /^\b[0-9a-fA-F]+\b$/;
+const MODULE = /^(?:[a-z ]+,)*[a-z ]+$/;
```
sameersubudhi, line 37: Please update the regex for MODULE name.
```suggestion
const MODULE = /^\b(?:[\w!@$&.]{1,32}|,)+\b$/;
```

framework/src/mysql.js
sameersubudhi: This update will induce a lot of changes within the entire codebase. Let's skip this for now. We can look into this later.

priojeetpriyom: We can just rename it here as atleast. When I was reading it, got an wrong idea thinking `property` is a key in the json. Later found it is actually the column name reading the knex doc

PR #1381, opened 2022-12-14
Implement reward sharing

### LINKED ISSUE #1380: Implement reward sharing
### Description

[LIP-0071](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0071.md) now introduces the dynamic reward sharing mechanism. Lisk Service must consider the validator `commission` and the actual block generation `reward` to compute and index the block generator's `totalCommission` and `totalSelfStakeRewards` values during the indexing process.

### Acceptance Criteria

- The indexing process applies each block sequentially, one at a time
- Indexing process starts indexing blocks in chronological order, starting with the genesis block
- The `commissions` table is updated with validator information for the genesis block at init (available in `pos` assets - data.validators)
- The `stakes` table is updated at init with relevant information from the block assets (`pos` - data.stakers)
- The block indexing process should compute the following values for the block generator and index them in the `validators` table only if `reward !== BigInt('0')`:
  - totalCommission
  - totalSelfStakeRewards

nagdahimanshu:
### What was the problem?
This PR resolves #1380 

### How was it solved?

- [x] The indexing process applies each block sequentially, one at a time
- [x] Indexing process starts indexing blocks in chronological order, starting with the genesis block (Already implemented, verified.)
- [x] The commissions table is updated with validator information for the genesis block at init (available in pos assets - data.validators)
- [x] The stakes table is updated at init with relevant information from the block assets (pos - data.stakers)
- [x] The block indexing process should compute the following values for the block generator and index them in the validators table only if reward !== BigInt('0'):
  - [x] totalCommission
  - [x] totalSelfStakeRewards

### How was it tested?
Local

### DIFF
--- a/services/blockchain-indexer/shared/indexer/blockchainIndex.js
+++ b/services/blockchain-indexer/shared/indexer/blockchainIndex.js
@@ -180,10 +180,10 @@ const indexBlock = async job => {
 			if (blockRewardEvent) {
 				blockReward = BigInt(blockRewardEvent.data.amount || '0');
 
-				if (blockReward) {
+				if (blockReward !== BigInt('0')) {
 					// TODO: Implement proper logic
-					const commission = calculateCommission(blockReward);
-					const selfStakeReward = calculateSelfStakeRewards(blockReward);
+					const commission = await calculateCommission(blockReward);
+					const selfStakeReward = await calculateSelfStakeRewards(block.generatorAddress);
 
 					await validatorsTable.increment({
 						increment: { totalCommission: BigInt(commission) },
--- a/services/blockchain-indexer/shared/indexer/blockchainIndex.js
+++ b/services/blockchain-indexer/shared/indexer/blockchainIndex.js
@@ -183,7 +183,11 @@ const indexBlock = async job => {
 				if (blockReward !== BigInt('0')) {
 					// TODO: Implement proper logic
 					const commission = await calcCommission(block.generatorAddress, blockReward);
-					const selfStakeReward = await calcSelfStakeReward(block.generatorAddress, blockReward);
+					const selfStakeReward = await calcSelfStakeReward(
+						block.generatorAddress,
+						blockReward,
+						commission,
+					);
 
 					await validatorsTable.increment({
 						increment: { totalCommission: BigInt(commission) },
--- a/services/blockchain-indexer/shared/indexer/blockchainIndex.js
+++ b/services/blockchain-indexer/shared/indexer/blockchainIndex.js
@@ -44,7 +44,7 @@ const {
 const { getLisk32AddressFromPublicKey, updateAccountPublicKey } = require('../utils/accountUtils');
 const { normalizeTransaction } = require('../utils/transactionsUtils');
 const { getEventsInfoToIndex } = require('../utils/eventsUtils');
-const { calculateCommission, calculateSelfStakeRewards } = require('../utils/validatorUtils');
+const { calcCommission, calcSelfStakeReward } = require('../utils/validatorUtils');
 
 const { updateAddressBalanceQueue } = require('./tokenIndex');
 
@@ -182,8 +182,8 @@ const indexBlock = async job => {
 
 				if (blockReward !== BigInt('0')) {
 					// TODO: Implement proper logic
-					const commission = await calculateCommission(block.generatorAddress, blockReward);
-					const selfStakeReward = await calculateSelfStakeRewards(block.generatorAddress);
+					const commission = await calcCommission(block.generatorAddress, blockReward);
+					const selfStakeReward = await calcSelfStakeReward(block.generatorAddress, blockReward);
 
 					await validatorsTable.increment({
 						increment: { totalCommission: BigInt(commission) },
--- a/services/blockchain-indexer/shared/indexer/blockchainIndex.js
+++ b/services/blockchain-indexer/shared/indexer/blockchainIndex.js
@@ -181,7 +181,6 @@ const indexBlock = async job => {
 				blockReward = BigInt(blockRewardEvent.data.amount || '0');
 
 				if (blockReward !== BigInt('0')) {
-					// TODO: Implement proper logic
 					const commission = await calcCommission(block.generatorAddress, blockReward);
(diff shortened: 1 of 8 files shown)


services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -198,18 +180,24 @@ const indexBlock = async job => {
...
-				await validatorsTable.increment({
-					increment: { totalCommission: commission },
-					where: { address: block.generatorAddress },
-				}, dbTrx);
-				await validatorsTable.increment({
-					increment: { totalSelfStakeRewards: selfStakeReward },
-					where: { address: block.generatorAddress },
-				}, dbTrx);
+				if (blockReward !== BigInt('0')) {
+					// TODO: Implement proper logic
```
ManuGowda, line 184: please remove

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -314,7 +302,7 @@ const deleteIndexedBlocks = async job => {
 };
 
 // Initialize queues
-const indexBlocksQueue = Queue(config.endpoints.cache, 'indexBlocksQueue', indexBlock, 30);
+const indexBlocksQueue = Queue(config.endpoints.cache, 'indexBlocksQueue', indexBlock, 1);
```
ManuGowda, line 305: I have my doubts here, it might introduce slowness, lets test and ensure the index performance.

services/blockchain-indexer/shared/utils/validatorUtils.js
```diff
@@ -20,15 +20,33 @@ const {
...
+	MYSQL_ENDPOINT,
 );
 
+const getStakesTable = () => getTableInstance(
+	stakesTableSchema.tableName,
+	stakesTableSchema,
+	MYSQL_ENDPOINT,
+);
+
+const MAX_COMMISSION = BigInt('10000');
```
ManuGowda, line 48: maybe this value should come from either SDK/service config/constant

ManuGowda (commented):
left few comments.

ManuGowda (approved):

services/blockchain-indexer/shared/utils/validatorUtils.js
```diff
@@ -51,7 +68,36 @@ const getAddressByName = async (name) => {
...
+	let selfStakeReward = BigInt('0');
+
+	const stakesTable = await getStakesTable();
+	const stakerInfo = await stakesTable.find(
+		{ validatorAddress: generatorAddress }, ['stakerAddress', 'amount'],
+	);
+
+	if (stakerInfo.length) {
+		const selfStakesInfo = stakerInfo.filter(stake => stake.stakerAddress === generatorAddress);
+		const selfStakes = selfStakesInfo.reduce((a, b) => BigInt(a.amount) + BigInt(b.amount));
```
gkoumout, line 89: I think this information could be get directly from some endpoint instead of being re-calculated here. 

In the PoS LIP we have the `getDelegate(address: Address)` endpoint (to be renamed `getValidator(address)` after terminology change) which returns all the information of the validator in the state store. 

The resulting object follows [this schema](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#json-schema-1) and contains `selfVotes` (to be renamed `selfStake`) and `totalVotesReceived` (to be renamed `totalStake`), so I guess you can get this values from there instead of scanning the whole stake database and adding the values, which increases the running time in the UI

services/blockchain-indexer/shared/utils/validatorUtils.js
```diff
@@ -51,7 +68,36 @@ const getAddressByName = async (name) => {
 	return null;
 };
 
+const calcCommission = async (generatorAddress, reward) => {
+	const commissionsTable = await getCommissionsTable();
+	const [{ commission: currentCommission }] = await commissionsTable
+		.find({ address: generatorAddress, sort: 'height:desc', limit: 1 }, 'commission');
+	const commission = (BigInt(reward) * BigInt(currentCommission)) / MAX_COMMISSION;
```
gkoumout, line 75: We have to make sure that the result of such floating number division is same as the one produced in the protocol implementation (which uses Q96 numbers)

gkoumout (changes requested):
Two comments from my side.

services/blockchain-indexer/shared/utils/validatorUtils.js
sameersubudhi: Assuming the current blockchain height is at height `hCurrent` and the highlighted calculation is made at height `h` which is lower than `hCurrent`. The issue with querying `getValdiator` is that the information that the node retrieves will be the updated state as of height `hCurrent` and not at height `h`. Hence, to compute the `totalSelfStake` and `totalCommission` values for every validator, Lisk Service requires the `totalStake` and `selfStake` values for the block generator at height `h` (lesser than equal to `hCurrent`). And, to calculate the `totalStake` and `selfStake` values at height `h`, we will need to be compute the `totalStake` and filter the `selfStake` values for the block generator at height `h`.

We can discuss this offline in detail, if the above comment is not clear enough.

sameersubudhi: @gkoumout Assuming the current blockchain height is at height `hCurrent` and the highlighted calculation is made at height `h` which is lower than `hCurrent`. The issue with querying `getValdiator` is that the information that the node retrieves will be the updated state as of height `hCurrent` and not at height `h`. Hence, to compute the `totalSelfStake` and `totalCommission` values for every validator, Lisk Service requires the `totalStake` and `selfStake` values for the block generator at height `h` (lesser than equal to `hCurrent`). And, to calculate the `totalStake` and `selfStake` values at height `h`, we will need to be compute the `totalStake` and filter the `selfStake` values for the block generator at height `h`.

We can discuss this offline in detail, if the above comment is not clear enough.

gkoumout: After discussing with Sameer and getting a better understanding of the explanation, can be resolved from my side.

gkoumout (approved):

PR #1404, opened 2023-01-05
Improper legacy cache handling

### LINKED ISSUE #1401: Improper legacy cache handling
### Actual behavior

Lisk Service when cold starts and applies the DB snapshot automatically, does not properly handle the legacy balance cache for already reclaimed addresses leading to poor user experience.

### Expected behavior

Legacy balance cache should be handled properly.

`Lisk Service v0.6.4`

sameersubudhi:
### What was the problem?
This PR resolves #1401 

### How was it solved?
- [x] Read legacy account info directly from the node
- [x] Remove legacy account info caching
- [x] Increase Redis maxmemory requirement to 1 GB

### How was it tested?
Local

### DIFF
--- a/docker/redis.persistent.conf
+++ b/docker/redis.persistent.conf
@@ -13,5 +13,5 @@ aof-use-rdb-preamble yes
 save ""
 
 # Memory management
-maxmemory 512mb
+maxmemory 1gb
 maxmemory-policy volatile-lru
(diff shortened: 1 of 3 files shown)


priojeetpriyom (approved):

nagdahimanshu (approved):
LGTM 👍

PR #1406, opened 2023-01-05
Update Lisk SDK to v6.0.0-alpha.10

### LINKED ISSUE #1405: Update Lisk SDK to v6.0.0-alpha.10
### Description

Update Lisk SDK to v6.0.0-alpha.10

### Acceptance Criteria

- Update SDK to v6.0.0-alpha.10

sameersubudhi:
### What was the problem?
This PR resolves #1405 

### How was it solved?

- [x] Update Lisk SDK to v6.0.0-alpha.10 dependencies on all microservices

### How was it tested?
Locally against `Lisk Core v4.0.0-alpha.8`, running on Lisk SDK v6.0.0-alpha.10


services/newsfeed/package.json
```diff
@@ -33,7 +33,7 @@
     "bluebird": "^3.7.2",
     "html-entities": "^2.3.2",
     "html-to-text": "^8.0.0",
-    "knex": "^2.0.0",
+    "knex": "^2.3.0",
```
nagdahimanshu, line 36: I think we can remove `knex` as a dependency as we are not using it directly. The current newsfeed implementation is directly using MySQL from framework.

Makefile
```diff
@@ -105,8 +105,8 @@ build-tests:
 	cd ./tests && docker build --tag=lisk/service_tests ./
 
 build-local:
-	npm ci
-	cd ./framework && npm ci
+	npm i --registry https://npm.lisk.com
+	cd ./framework && npm i --registry https://npm.lisk.com
```
priojeetpriyom, line 109: Could using `npm install` create an issue for cases when our code becomes incompatible with new libraries?

services/newsfeed/package.json
sameersubudhi: Addressed!

Makefile
sameersubudhi: This change is temporary to bypass current dependency installation issues. This won't make it to the production.

priojeetpriyom (approved):

nagdahimanshu (approved):

PR #1452, opened 2023-01-18
Update README

### LINKED ISSUE #1451: Update README
### Description

Since we do not publish lisk-service-framework to the npm registry, the README should reflect the same

### Acceptance Criteria

- Remove irrelevant information from [README](https://github.com/LiskHQ/lisk-service/tree/development/framework#installation)

sameersubudhi:
### What was the problem?
This PR resolves #1451 

### How was it solved?
- [x] Update README with proper instructions to install `lisk-service-framework` as a dependency

### How was it tested?
- Executed the command locally

### DIFF
--- a/framework/README.md
+++ b/framework/README.md
@@ -12,8 +12,9 @@ All modules can be easily replaced by other libraries when needed.
 
 ## Installation
 
+Set the target URL to the desired tarball distribution permalink.
 ```
-npm install lisk-service-framework
+npm install https://github.com/LiskHQ/lisk-service/raw/2fbc65f49ae5c5aaa3689667051dbd4082b8c25a/framework/dist/lisk-service-framework-1.4.10.tgz
 ```
 
 ## Quick Start


framework/README.md
```diff
@@ -12,8 +12,9 @@ All modules can be easily replaced by other libraries when needed.
 
 ## Installation
 
+Set the target URL to the desired tarball distribution permalink.
 ```
-npm install lisk-service-framework
+npm install https://github.com/LiskHQ/lisk-service/raw/2fbc65f49ae5c5aaa3689667051dbd4082b8c25a/framework/dist/lisk-service-framework-1.4.10.tgz
```
ManuGowda, line 17: is this how we currently doing it?

sameersubudhi: Yes!

ManuGowda (approved):

framework/README.md
ManuGowda: Probably every-time we update the version we have remember to update this version i guess. 
we need to find a better approach, probably if we need to publish the `lisk-service-framework` we should consider doing that in this release?

sameersubudhi: Everytime we make changes to the framework, we package the distribution and check it in the repo itself under the path: `./framework/dist`

PR #1480, opened 2023-02-01
Add integration tests

### LINKED ISSUE #1466: Add integration tests
### Description

Integration tests are necessary to test and ensure proper e2e response.
These ensure proper response structure and sometimes ensure proper implementation of the application logic.

### Acceptance Criteria

- Integrations tests are added/updated for the following endpoints (HTTP and RPC)
  - `/invoke`
  - `/spec`
  - `/ready`
  - `/status`

nagdahimanshu:
### What was the problem?
This PR resolves #1466 

### How was it solved?

- [x] Integrations tests are added/updated for the following endpoints (HTTP and RPC)
  - [x] /invoke
  - [x] /spec
  - [x] /ready
  - [x] /status

### How was it tested?
Local


tests/schemas/api_v3/regex.js
```diff
@@ -40,6 +40,7 @@ const HEX = /^\b[0-9a-fA-F]+\b$/;
 const TOPIC = /^\b(?:[0-9a-fA-F]+|lsk[a-hjkm-z2-9]{38})\b$/;
 const EMPTY_STRING = /^$/;
 const POS_VALIDATOR_STATUS = /^\b(?:active|standby|banned|punished|ineligible|,)+\b$/;
+const SWAGGER_SCHEMES = /^\b(?:http|https){1}\b$/;
```
sameersubudhi, line 43: Can be shortened to, but the existing also looks fine.
```suggestion
const SWAGGER_SCHEMES = /^https?$/;
```

tests/integration/gateway/ready.test.js
```diff
@@ -0,0 +1,42 @@
...
+} = require('../../schemas/gateway.schema');
+
+const baseUrl = config.SERVICE_ENDPOINT;
+const endpoint = `${baseUrl}/api/ready`;
+
+describe('Ready API', () => {
+	describe(`GET ${endpoint}`, () => {
+		it('Report readiness -> 200 OK', async () => {
+			const response = await api.get(endpoint);
+			expect(response).toMap(readySchema);
```
sameersubudhi, line 34: Please include the fallback scenario for when the microservices are not ready, probably wrapped within a `catch` block!?

tests/integration/api_v3/http/spec.test.js
```diff
@@ -0,0 +1,41 @@
...
+} = require('../../../schemas/api_v3/spec.schema');
+
+const baseUrl = config.SERVICE_ENDPOINT;
+const baseUrlV3 = `${baseUrl}/api/v3`;
+const endpoint = `${baseUrlV3}/spec`;
+
+describe('Swagger spec API', () => {
+	it('return swagger specifications -> 200 OK', async () => {
+		const response = await api.get(endpoint);
+		expect(response).toMap(specResponseSchema);
```
sameersubudhi, line 34: The test case can be expanded to be a bit more exhaustive? Alternatively, the schema can also be adjusted to make the checks a little more strict. We can discuss this offline.

sameersubudhi (changes requested):
Looks good overall. But please consider the following.

tests/schemas/api_v3/invoke.schema.js
```diff
@@ -0,0 +1,35 @@
...
+import Joi from 'joi';
+
+const invokeMetaResponseSchema = {
+	endpoint: Joi.string().required(),
+	params: Joi.object().optional(),
+};
+
+const errorSchema = {
+	message: Joi.string().required(),
+};
```
priojeetpriyom, line 25: would it be better to put it in http generics instead?

tests/schemas/gateway.schema.js
```diff
@@ -14,25 +14,25 @@
...
-	transactionStatsStatus: Joi.boolean().required(),
-	feesStatus: Joi.boolean().required(),
-	delegatesStatus: Joi.boolean().required(),
+	indexer: Joi.boolean().required(),
+	connector: Joi.boolean().required(),
+	fees: Joi.boolean().optional(),
+	market: Joi.boolean().optional(),
+	newsfeed: Joi.boolean().optional(),
+	'app-registry': Joi.boolean().optional(),
+	statistics: Joi.boolean().optional(),
```
priojeetpriyom, line 35: would making gateway a folder and moving the two schemas under two files make it more consistent with the current structure?

tests/schemas/api_v3/invoke.schema.js
nagdahimanshu: This schema is more specific to the invoke endpoint response. We can clarify offline if needed.

tests/integration/gateway/ready.test.js
```diff
@@ -0,0 +1,49 @@
...
+	describe(`GET ${endpoint}`, () => {
+		it('Report readiness -> 200 OK', async () => {
+			try {
+				const response = await api.get(endpoint);
+				expect(response).toMap(readySchema);
+			} catch (_) {
+				const expectedResponseCode = 503;
+				const response = await api.get(endpoint, expectedResponseCode);
+				expect(response).toMap(serviceUnavailableSchema);
+			}
```
priojeetpriyom, line 41: Can't we only test the following case like we do for all other tests? it will also cover the test in try.
```suggestion
			const response = await api.get(endpoint, 503);
			expect(response).toMap(serviceUnavailableSchema);
```

priojeetpriyom (approved):

tests/integration/gateway/ready.test.js
sameersubudhi: No, the idea here is to check for the successful response within the `try` block and in case the `/ready` endpoint is not ready it responds with a `503 SERVICE UNAVAILABLE`, which happens to be vet in the `catch` block.

If you only check for a `503 SERVICE UNAVAILABLE`, then the tests would fail if the `/ready` endpoint returns with a `200 OK` response.

sameersubudhi (approved):

priojeetpriyom (approved):

tests/integration/gateway/ready.test.js
sameersubudhi: This should be fine for now, as is.

Issue #1391, opened 2022-12-28
Discrepancy between totalProduced and total generated blocks


przemerr:
### Actual behavior
`totalProduced` in https://service.lisk.com/api/v2/accounts?username=friartuck  doesn't equal 
`total` in https://service.lisk.com/api/v2/blocks?generatorUsername=friartuck&limit=1 

### Expected behavior

The values should be equal.

### Steps to reproduce
Make the above queries and check the responses
### Which version(s) does this affect? (Environment, OS, etc...)
0.6.4 official mainnet

Issue #1389, opened 2022-12-28
Update Makefile


nagdahimanshu:
### Description

Update Makefile and Jenkinsfile with the latest changes

Acceptance Criteria
- Makefile is up-to-date
- Update `build-local` command to use `npm ci`
- Update `Jenkinsfile` Build dependencies section to use `npm ci`
- All commands should work as expected

Issue #1387, opened 2022-12-19
Incorrect delegate consecutiveMissedBlocks count


sameersubudhi:
### Actual behavior
The `consecutiveMissedBlocks` for the following delegates are shown incorrectly:

https://service.lisk.com/api/v2/accounts?username=condor
https://service.lisk.com/api/v2/accounts?username=coinc

### Expected behavior

The `consecutiveMissedBlocks` for all the delegates must be shown correctly.

### Steps to reproduce
Check the response for the above delegates on Lisk Service (https://service.lisk.com/api/v2/accounts?username=<username>) and on Lisk Core (https://mainnet.lisk.com/api/accounts/<hexAddress>).

### Which version(s) does this affect? (Environment, OS, etc...)
Lisk Service `v0.6.4`

przemerr:
I'm adding additional delegate https://service.lisk.com/api/v2/accounts?username=savetheworld because delegates can go back to being online resetting the count. This one should stick indefinitely since he's on a verge of a ban.

przemerr:
If you compare https://service.lisk.com/api/v2/accounts?username=savetheworld with https://mainnet.lisk.com/api/accounts/0b987a9b1b3b106ba709deb5f88290f9b200a866
not only `consecutiveMissedBlocks` are wrong but also the `isBanned` flag, the delegate got banned recently (few hours ago).

Issue #1264, opened 2022-09-15
Fix minor discrepancy with APIv2 swagger specs


nagdahimanshu:
### Actual behavior
Swagger API does not allow user to query transaction with valid transactionID

### Expected behavior
It should allow user to query transactions by all the supported query params.

### Steps to reproduce
- Run gateway service and call endpoint `/api/v2/spec`
- validate the response with Swagger editor

### Which version(s) does this affect? (Environment, OS, etc...)
V0.6.4

PR #1493, opened 2023-02-08
Cherry-pick fixes for v0.6.5 release

### LINKED ISSUE #1487: Cherry-pick fixes for v0.6.5 release
### Description

Prepare for the `v0.6.5` release.

### Acceptance Criteria

- Dependency vulnerabilities are fixed
  - CVE-2016-20018
- All microservices are up-to-date with the dependency updates
- The following issues are addressed
  - #1264
  - #1389 
  - #1391 
  - #1387 
  - #1229

sameersubudhi:
### What was the problem?
This PR resolves #1487 

### How was it solved?

- [x] Dependency vulnerabilities are fixed
  - CVE-2016-20018
- [x] All microservices are up-to-date with the dependency updates
- [x] Cherry-picked all the commits resolving the following issues:
  - #1264
  - #1389 
  - #1391 
  - #1387 
  - #1229  

### How was it tested?
- Multiple deployments against Mainnet and Testnet
  - Ran integration tests for consistency checks
  - Manual sanity checks for the reported errors


docker/redis.persistent.conf
```diff
@@ -13,5 +13,5 @@ aof-use-rdb-preamble yes
 save ""
 
 # Memory management
-maxmemory 512mb
+maxmemory 2gb
```
ManuGowda, line 16: Thats a drastic increase 4 times, probably the docs or some place this require a update?

framework/bin/socket_io_subscribe_client.js
```diff
@@ -46,6 +46,7 @@ const subscribe = event => {
 	});
 };
 
+// subscribe('*'); // Listen to all the events
```
ManuGowda, line 49: Is this change intended for this patch?

framework/src/mysql.js
```diff
@@ -13,6 +13,8 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+// TODO: Rename all methods to use 'DB' instead of 'Db'
```
ManuGowda, line 16: TODO should be resolved?

framework/src/mysql.js
```diff
@@ -281,6 +306,7 @@ const getTableInstance = async (tableName, tableConfig, connEndpoint = CONN_ENDP
 	};
 
 	const find = async (params = {}, columns) => {
+		// TODO: Add support for IS NULL, IS NOT NULL, LIKE
```
ManuGowda, line 309: same here

services/core/shared/core/compat/sdk_v5/accounts.js
```diff
@@ -133,7 +133,10 @@ const getAccountsFromCore = async (params) => {
 
 		await BluebirdPromise.map(
 			accounts.data,
-			async account => accountsCache.set(account.address, JSON.stringify(account)),
+			async account => {
+				const ttlMillisecs = 5 * 60 * 1000; // Expire cache entry in 5 mins
```
ManuGowda, line 137: do you think it would be better to get this value from config? in case if we need to increase or decrease we can ask devops to change and restart

ManuGowda (changes requested):
Why do we have the dist files pushed 🤔 e.g framework/dist/lisk-service-framework-1.4.10.tgz

docker/redis.persistent.conf
sameersubudhi: IIRC, we actually use `4gb` in production. It made more sense to raise it up to a more realistic level given the size of the mainnet.

framework/src/mysql.js
sameersubudhi: No, we'll be resolving it as part of the `0.7.0` release.

framework/src/mysql.js
sameersubudhi: Same as above.

services/core/shared/core/compat/sdk_v5/accounts.js
sameersubudhi: Sure, I'll add it as a config with a supporting environment variable.

sameersubudhi:
> Why do we have the dist files pushed 🤔 e.g framework/dist/lisk-service-framework-1.4.10.tgz

This is how we have been publishing the lisk-service-framework. Ideally, it should've been from the npm registry.

services/core/shared/core/compat/sdk_v5/accounts.js
sameersubudhi: Done!

ManuGowda:
> > Why do we have the dist files pushed 🤔 e.g framework/dist/lisk-service-framework-1.4.10.tgz
> 
> This is how we have been publishing the lisk-service-framework. Ideally, it should've been from the npm registry.

Ok let's then create an issue on GitHub so that we can address it 🙏🏼

ManuGowda (approved):

PR #1505, opened 2023-02-14
Update totalLocked on block delete

### LINKED ISSUE #1440: Update totalLocked on block delete
### Description

The `totalLocked` is aggregated within the Lisk Service Indexer microservice based on events.
However, the events in case of delete block events cannot be queried from the node. Hence, Lisk Service needs to cache the events corresponding all the non-finalized blocks and update them as and when the block finality is updated.

### Acceptance Criteria

- Cache events corresponding the non-final blocks within the indexer
  - Consider storing all the events for the non-final blocks in MySQL (no matter the value for `config.db.isPersistEvents`) and delete them once the blocks are finalized based on the `config.db.isPersistEvents` value
- Introduce a new method: `getCachedEventsByBlockID` that reads and serves the information from the cache in indexer
- Implement the necessary changes for `totalLocked` within the delete block event handler in indexer

priojeetpriyom:
### What was the problem?
This PR resolves #1440 

### How was it solved?

- [x] Cache events corresponding the non-final blocks within the indexer
- [x] Consider storing all the events for the non-final blocks in MySQL (no matter the value for config.db.isPersistEvents) and delete them once the blocks are finalized based on the config.db.isPersistEvents value
- [x] Introduce a new method: getCachedEventsByBlockID that reads and serves the information from the cache in indexer
- [x] Implement the necessary changes for totalLocked within the delete block event handler in indexer

### How was it tested?
- [x] Local

### DIFF
--- a/services/blockchain-connector/shared/sdk/network.js
+++ b/services/blockchain-connector/shared/sdk/network.js
@@ -39,6 +39,11 @@ const refreshNetworkStatus = async () => {
 			logger.warn(`Error occurred while refreshing network status:\n${err.stack}`);
 		}
 	};
+
+	if (!networkStatus) {
+		await refreshNetworkStatusListener();
+	}
+
 	Signals.get('chainNewBlock').add(refreshNetworkStatusListener);
 };
 
--- a/services/blockchain-indexer/shared/dataService/business/events.js
+++ b/services/blockchain-indexer/shared/dataService/business/events.js
@@ -79,6 +79,8 @@ const getEventsByHeight = async (height) => {
 	return eventsFromNode;
 };
 
+const deleteEventsFromCache = async (height) => eventCache.delete(height);
+
 const getEvents = async (params) => {
 	const blocksTable = await getBlocksTable();
 	const eventsTable = await getEventsTable();
@@ -186,4 +188,5 @@ const getEvents = async (params) => {
 module.exports = {
 	getEvents,
 	getEventsByHeight,
+	deleteEventsFromCache,
 };
--- a/services/blockchain-indexer/shared/dataService/business/events.js
+++ b/services/blockchain-indexer/shared/dataService/business/events.js
@@ -25,9 +25,9 @@ const {
 
 const config = require('../../../config');
 
-const blocksIndexSchema = require('../../database/schema/blocks');
-const eventsIndexSchema = require('../../database/schema/events');
-const eventTopicsIndexSchema = require('../../database/schema/eventTopics');
+const blocksTableSchema = require('../../database/schema/blocks');
+const eventsTableSchema = require('../../database/schema/events');
+const eventTopicsTableSchema = require('../../database/schema/eventTopics');
 
 const { requestConnector } = require('../../utils/request');
 const { normalizeRangeParam } = require('../../utils/paramUtils');
@@ -35,19 +35,19 @@ const { parseToJSONCompatObj } = require('../../utils/parser');
 
 const MYSQL_ENDPOINT = config.endpoints.mysql;
 
-const getBlocksIndex = () => getTableInstance(
-	blocksIndexSchema.tableName,
-	blocksIndexSchema,
+const getBlocksTable = () => getTableInstance(
+	blocksTableSchema.tableName,
+	blocksTableSchema,
 	MYSQL_ENDPOINT,
 );
-const getEventsIndex = () => getTableInstance(
-	eventsIndexSchema.tableName,
-	eventsIndexSchema,
+const getEventsTable = () => getTableInstance(
+	eventsTableSchema.tableName,
+	eventsTableSchema,
 	MYSQL_ENDPOINT,
 );
-const getEventTopicsIndex = () => getTableInstance(
-	eventTopicsIndexSchema.tableName,
-	eventTopicsIndexSchema,
+const getEventTopicsTable = () => getTableInstance(
+	eventTopicsTableSchema.tableName,
+	eventTopicsTableSchema,
 	MYSQL_ENDPOINT,
 );
 
@@ -59,19 +59,30 @@ const getEventsByHeight = async (height) => {
 };
 
 const getEventsFromCache = async (height) => {
-	const events = await eventCache.get(height);
-	if (!events) {
-		const eventFromNode = await getEventsByHeight(height);
-		await eventCache.set(height, JSON.stringify(eventFromNode));
-		return eventFromNode;
+	// Get from cache
+	const cacheEvents = await eventCache.get(height);
+	if (cacheEvents) return JSON.parse(cacheEvents);
+
+	// Get from db
+	const eventsTable = await getEventsTable();
+	const dbEventStrs = await eventsTable.find({ height }, ['eventStr']);
+
+	if (dbEventStrs.length) {
+		const dbEvents = dbEventStrs.map(({ eventStr }) => eventStr ? JSON.parse(eventStr) : eventStr);
+		await eventCache.set(height, JSON.stringify(dbEvents));
+		return dbEvents;
 	}
-	return JSON.parse(events);
+
+	// Get from node
+	const eventsFromNode = await getEventsByHeight(height);
+	await eventCache.set(height, JSON.stringify(eventsFromNode));
+	return eventsFromNode;
 };
 
 const getEvents = async (params) => {
-	const blocksTable = await getBlocksIndex();
-	const eventsTable = await getEventsIndex();
-	const eventTopicsTable = await getEventTopicsIndex();
+	const blocksTable = await getBlocksTable();
+	const eventsTable = await getEventsTable();
+	const eventTopicsTable = await getEventTopicsTable();
 
 	const events = {
 		data: [],
@@ -175,4 +186,5 @@ const getEvents = async (params) => {
 module.exports = {
 	getEvents,
 	getEventsByHeight,
+	getEventsFromCache,
 };
--- a/services/blockchain-indexer/shared/dataService/business/events.js
+++ b/services/blockchain-indexer/shared/dataService/business/events.js
@@ -53,12 +53,12 @@ const getEventTopicsTable = () => getTableInstance(
 
 const eventCache = CacheLRU('events');
 
-const getEventsByHeight = async (height) => {
+const getEventsByHeightFromNode = async (height) => {
 	const events = await requestConnector('getEventsByHeight', { height });
 	return parseToJSONCompatObj(events);
 };
 
-const getEventsFromCache = async (height) => {
+const getEventsByHeight = async (height) => {
 	// Get from cache
 	const cachedEvents = await eventCache.get(height);
 	if (cachedEvents) return JSON.parse(cachedEvents);
@@ -74,7 +74,7 @@ const getEventsFromCache = async (height) => {
 	}
 
 	// Get from node
-	const eventsFromNode = await getEventsByHeight(height);
+	const eventsFromNode = await getEventsByHeightFromNode(height);
 	await eventCache.set(height, JSON.stringify(eventsFromNode));
 	return eventsFromNode;
 };
@@ -158,7 +158,7 @@ const getEvents = async (params) => {
 				if (eventStr) event = JSON.parse(eventStr);
 			}
 			if (!event) {
-				const eventsFromCache = await getEventsFromCache(height);
+				const eventsFromCache = await getEventsByHeight(height);
 				event = eventsFromCache.find(entry => entry.index === index);
 			}
 
@@ -186,5 +186,4 @@ const getEvents = async (params) => {
 module.exports = {
 	getEvents,
 	getEventsByHeight,
-	getEventsFromCache,
 };
--- a/services/blockchain-indexer/shared/dataService/business/events.js
+++ b/services/blockchain-indexer/shared/dataService/business/events.js
@@ -60,10 +60,10 @@ const getEventsByHeight = async (height) => {
 
 const getEventsFromCache = async (height) => {
 	// Get from cache
-	const cacheEvents = await eventCache.get(height);
-	if (cacheEvents) return JSON.parse(cacheEvents);
+	const cachedEvents = await eventCache.get(height);
+	if (cachedEvents) return JSON.parse(cachedEvents);
 
-	// Get from db
+	// Get from DB
 	const eventsTable = await getEventsTable();
 	const dbEventStrs = await eventsTable.find({ height }, ['eventStr']);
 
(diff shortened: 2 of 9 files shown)


services/blockchain-connector/shared/sdk/network.js
```diff
@@ -39,6 +39,11 @@ const refreshNetworkStatus = async () => {
 			logger.warn(`Error occurred while refreshing network status:\n${err.stack}`);
 		}
 	};
+
+	if (!networkStatus) {
+		await refreshNetworkStatusListener();
+	}
```
nagdahimanshu, line 45: I think this part is not correct, We need to update network status on every new block even if already exists in-memory. There are some properties which is getting updated on every new block, for e.g. `finalizedHeight`
```suggestion
	if (!networkStatus) {
		await refreshNetworkStatusListener();
	}
```

sameersubudhi: This is executed only once at init. Required, in case the cache isn't set before the event listener kicks in and the indexer is querying for the network status.

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -111,24 +119,19 @@ const getTransactionExecutionStatus = (tx, events) => {
...
-// 		await keyValueTable.set(tokenKey, newLockedAmount, dbTrx);
-// 	},
-// 	{ concurrency: Object.entries(tokenIDLockedAmountChangeMap).length },
-// );
+const updateTotalLockedAmounts = (tokenIDLockedAmountChangeMap, dbTrx) => BluebirdPromise.map(
+	Object.entries(tokenIDLockedAmountChangeMap),
+	async ([tokenID, lockedAmountChange]) => {
+		if (lockedAmountChange === BigInt(0)) return;
+
+		const tokenKey = KV_STORE_KEY.PREFIX.TOTAL_LOCKED.concat(tokenID);
```
sameersubudhi, line 127: Please set the value to `0` instead. Else, when querying, the response will be undefined/null.
```suggestion
	async ([tokenID, lockedAmountChange]) => {
		const tokenKey = KV_STORE_KEY.PREFIX.TOTAL_LOCKED.concat(tokenID);
```

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -315,29 +323,27 @@ const deleteIndexedBlocks = async job => {
...
-				// 	// Negate amount to reverse the affect
-				// 	if (event.name === EVENT_NAME.LOCKED) {
-				// 		tokenIDLockedAmountChangeMap[eventData.tokenID] -= eventData.amount;
-				// 	} else if (event.name === EVENT_NAME.UNLOCKED) {
-				// 		tokenIDLockedAmountChangeMap[eventData.tokenID] += eventData.amount;
-				// 	}
-				// });
-				// await updateTotalLockedAmounts(tokenIDLockedAmountChangeMap, dbTrx);
-				// }
+				const events = await getEventsFromCache(block.height);
```
sameersubudhi, line 326: Is it possible to fetch the events based on the block ID instead? 🤔

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -315,29 +323,27 @@ const deleteIndexedBlocks = async job => {
...
+					const tokenIDLockedAmountChangeMap = {};
+					events.forEach(event => {
+						const { data: eventData } = event;
+						// Initialize map entry with BigInt
+						if ([EVENT_NAME.LOCK, EVENT_NAME.UNLOCK].includes(event.name)
+						&& !(eventData.tokenID in tokenIDLockedAmountChangeMap)) {
+							tokenIDLockedAmountChangeMap[eventData.tokenID] = BigInt(0);
+						}
+
+						// Negate amount to reverse the affect
```
sameersubudhi, line 338:
```suggestion
						// Negate amount to reverse the effect
```

sameersubudhi (changes requested):

services/blockchain-connector/shared/sdk/network.js
nagdahimanshu: Thanks for the clarification 👍

services/blockchain-indexer/shared/indexer/blockchainIndex.js
priojeetpriyom: Applied.

services/blockchain-indexer/shared/indexer/blockchainIndex.js
priojeetpriyom: Applied.

services/blockchain-indexer/shared/utils/eventsUtils.js
```diff
@@ -52,6 +74,23 @@ const getEventsInfoToIndex = async (block, events) => {
 	return eventsInfoToIndex;
 };
 
+const deleteEventsTillBlockHeight = async (blockHeight, dbTrx) => {
+	if (blockHeight === undefined || Number.isNaN(blockHeight)) return;
```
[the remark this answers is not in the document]
priojeetpriyom: Applied.

services/blockchain-indexer/shared/dataService/business/events.js
```diff
@@ -59,19 +59,30 @@ const getEventsByHeight = async (height) => {
...
-	const events = await eventCache.get(height);
-	if (!events) {
-		const eventFromNode = await getEventsByHeight(height);
-		await eventCache.set(height, JSON.stringify(eventFromNode));
-		return eventFromNode;
+	// Get from cache
+	const cacheEvents = await eventCache.get(height);
+	if (cacheEvents) return JSON.parse(cacheEvents);
+
+	// Get from db
```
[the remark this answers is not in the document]
priojeetpriyom: Applied.

services/blockchain-indexer/shared/utils/eventsUtils.js
```diff
@@ -52,6 +74,23 @@ const getEventsInfoToIndex = async (block, events) => {
...
+
+	const eventsTable = await getEventsTable();
+	const eventTopicsTable = await getEventTopicsTable();
+	const queryParams = {
+		propBetweens: [{
+			property: 'height',
+			lowerThan: blockHeight + 1,
```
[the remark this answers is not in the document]
priojeetpriyom: Actually lowerThan results in a similar if not better bound than to.
https://github.com/LiskHQ/lisk-service/blob/1440-update-total-locked-on-block-delete/framework/src/mysql.js#L242-L244

services/blockchain-indexer/shared/indexer/blockchainIndex.js
priojeetpriyom: Can be done, but would required refactoring of current access pattern and implementation. Also, could not find sdk endpoint to retrieve events by block id, currently we are using `chain_getEvents` which gets by block height

services/blockchain-indexer/shared/utils/eventsUtils.js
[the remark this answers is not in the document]
priojeetpriyom: Applied.

services/blockchain-indexer/shared/dataService/business/events.js
```diff
@@ -59,19 +59,30 @@ const getEventsByHeight = async (height) => {
 };
 
 const getEventsFromCache = async (height) => {
```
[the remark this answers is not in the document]
priojeetpriyom: Applied.

services/blockchain-indexer/shared/utils/eventsUtils.js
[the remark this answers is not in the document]
priojeetpriyom: The reason for this check is to avoid unnecessary query if blockHeight NaN

services/blockchain-indexer/shared/utils/eventsUtils.js
```diff
@@ -52,6 +74,23 @@ const getEventsInfoToIndex = async (block, events) => {
 	return eventsInfoToIndex;
 };
 
+const deleteEventsTillBlockHeight = async (blockHeight, dbTrx) => {
+	if (Number.isNaN(Number(blockHeight))) return;
```
sameersubudhi, line 78: This check isn't required but if left should be like below, since `Number.isNaN(Number(null))` will result in `false`.
```suggestion
	if (typeof blockHeight !== 'number') return;
```

services/blockchain-indexer/shared/utils/eventsUtils.js
```diff
@@ -52,6 +74,23 @@ const getEventsInfoToIndex = async (block, events) => {
...
+
+	const eventsTable = await getEventsTable();
+	const eventTopicsTable = await getEventTopicsTable();
+	const queryParams = {
+		propBetweens: [{
+			property: 'height',
+			to: blockHeight,
+		}],
+		limit: 10000,
```
sameersubudhi, line 87: Why the `limit`? What if there were more than 10000 eligible entries?
```suggestion
```

sameersubudhi (changes requested):

services/blockchain-indexer/shared/utils/eventsUtils.js
priojeetpriyom: those will be deleted eventually on calls for each next block. the reason to add limit is the following case - suppose blocks were indexed with `db.config.db.isPersistEvents = true` before and now the user set `config.db.isPersistEvents = false` and restarted indexer. Now this call will try to delete all the events and event topics of finalized blocks stored so far which can be a expensive operation. this limit will ease the load by cleaning up in chunks for each subsequent blocks.

services/blockchain-indexer/shared/utils/eventsUtils.js
priojeetpriyom: Removed.

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -43,7 +44,7 @@ const {
 
 const { getLisk32AddressFromPublicKey, updateAccountPublicKey } = require('../utils/accountUtils');
 const { normalizeTransaction } = require('../utils/transactionsUtils');
-const { getEventsInfoToIndex } = require('../utils/eventsUtils');
+const { getEventsInfoToIndex, deleteEventsTillBlockHeight } = require('../utils/eventsUtils');
```
sameersubudhi, line 47:
```suggestion
const { getEventsInfoToIndex, deleteEventsTillHeight } = require('../utils/eventsUtils');
```

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -218,24 +219,29 @@ const indexBlock = async job => {
...
+				if (event.name === EVENT_NAME.LOCK) {
+					tokenIDLockedAmountChangeMap[eventData.tokenID] += BigInt(eventData.amount);
+				} else if (event.name === EVENT_NAME.UNLOCK) {
+					tokenIDLockedAmountChangeMap[eventData.tokenID] -= BigInt(eventData.amount);
+				}
+			});
+			await updateTotalLockedAmounts(tokenIDLockedAmountChangeMap, dbTrx);
+		}
+
+		// Delete events of finalized blocks unless required to persist
```
sameersubudhi, line 241:
```suggestion
		// Delete events from the finalized blocks when persistence is disabled
```

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -315,29 +321,31 @@ const deleteIndexedBlocks = async job => {
...
+							tokenIDLockedAmountChangeMap[eventData.tokenID] -= BigInt(eventData.amount);
+						} else if (event.name === EVENT_NAME.UNLOCK) {
+							tokenIDLockedAmountChangeMap[eventData.tokenID] += BigInt(eventData.amount);
+						}
+					});
+					await updateTotalLockedAmounts(tokenIDLockedAmountChangeMap, dbTrx);
+				}
+
+				// Invalidate cached events of this block
+				// This must be done after processing all event related calculations
```
sameersubudhi, line 347:
```suggestion
				// Invalidate cached events for the current block
				// Must be done after the events are processed
```

services/blockchain-indexer/shared/utils/eventsUtils.js
```diff
@@ -52,6 +74,21 @@ const getEventsInfoToIndex = async (block, events) => {
 	return eventsInfoToIndex;
 };
 
+const deleteEventsTillBlockHeight = async (blockHeight, dbTrx) => {
```
sameersubudhi, line 77:
```suggestion
const deleteEventsTillHeight = async (blockHeight, dbTrx) => {
```

services/blockchain-indexer/shared/utils/eventsUtils.js
```diff
@@ -52,6 +74,21 @@ const getEventsInfoToIndex = async (block, events) => {
...
+		}],
+		limit: 10000,
+	};
+	await eventTopicsTable.delete(queryParams, dbTrx);
+	await eventsTable.delete(queryParams, dbTrx);
+};
+
 module.exports = {
 	getEventsInfoToIndex,
+	deleteEventsTillBlockHeight,
```
sameersubudhi, line 93:
```suggestion
	deleteEventsTillHeight,
```

sameersubudhi (approved):
Looks fine overall. Please consider the following recommendations.

services/blockchain-indexer/shared/utils/eventsUtils.js
priojeetpriyom: Applied.

services/blockchain-indexer/shared/utils/eventsUtils.js
priojeetpriyom: Applied.

services/blockchain-indexer/shared/indexer/blockchainIndex.js
priojeetpriyom: I would prefer keeping the previous comment as both are similar.

services/blockchain-indexer/shared/indexer/blockchainIndex.js
priojeetpriyom: Applied partially.

services/blockchain-indexer/shared/indexer/blockchainIndex.js
priojeetpriyom: Applied.

nagdahimanshu (approved):
LGTM 👍

Issue #1510, opened 2023-02-16
Fix chainID for peers response


sameersubudhi:
### Actual behavior

`chainID` for peers returns `"[object Object]"`.

```
{
  "data": [
    {
      "ip": "128.199.13.251",
      "port": 4001,
      "networkVersion": "1.0",
      "state": "connected",
      "height": 8157,
      "chainID": "[object Object]"
    },
    {
      "ip": "167.99.191.96",
      "port": 4001,
      "networkVersion": "1.0",
      "state": "connected",
      "height": 8157,
      "chainID": "[object Object]"
    }
  ],
  "meta": {
    "count": 2,
    "offset": 0,
    "total": 2
  }
}
```

### Expected behavior

`chainID` for must be `"03000000"` instead.

### Steps to reproduce

- Create a network and call the `/api/v3/peers` endpoint

### Which version(s) does this affect? (Environment, OS, etc...)

Lisk Service v0.7.0-alpha.7

PR #1514, opened 2023-02-21
Fix chainID for peers response

### LINKED ISSUE #1506: Fix example on swagger
### Actual behavior

The swagger example for `/blockchain/apps` endpoint display `lastUpdated` as a string instead of a integer.

```
{
  "data": [
    {
      "name": "Lisk",
      "chainID": "00000000",
      "state": "active",
      "address": "lskdwsyfmcko6mcd357446yatromr9vzgu7eb8y99",
      "lastCertificateHeight": 160,
      "lastUpdated": "1616008148"
    }
  ],
  "meta": {
    "count": 10,
    "offset": 10,
    "total": 400
  }
}
```

### Expected behavior

Last updated must be an integer both in the example and response.

### Steps to reproduce

- Run Lisk Service
- Query `/api/v3/spec` endpoint
- Check the response on `https://editor.swagger.io/`

Lisk Service v0.7.0-alpha.6

nagdahimanshu:
### What was the problem?
This PR resolves #1506  #1510

### How was it solved?

- [ ] Fix swagger example for `blockchain/apps`
- [ ] Fix `ChainID` mapping in connector (Parse to buffer to string) 

### How was it tested?
Local

### DIFF
--- a/services/blockchain-connector/shared/sdk/peerCache.js
+++ b/services/blockchain-connector/shared/sdk/peerCache.js
@@ -14,8 +14,9 @@
  *
  */
 const { Logger, Signals } = require('lisk-service-framework');
-const actions = require('./endpoints');
+const endpoints = require('./endpoints');
 const GeoService = require('../geolocation');
+const { parseToJSONCompatObj } = require('../utils/parser');
 
 const logger = Logger();
 
@@ -45,7 +46,7 @@ const refactorPeer = (orgPeer, state) => {
 	peer.state = state;
 	peer.height = height;
 	peer.ip = ipAddress;
-	return peer;
+	return parseToJSONCompatObj(peer);
 };
 
 const addLocation = async (ipaddress) => {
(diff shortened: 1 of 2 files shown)


services/blockchain-connector/shared/sdk/peerCache.js
```diff
@@ -45,7 +46,7 @@ const refactorPeer = (orgPeer, state) => {
 	peer.state = state;
 	peer.height = height;
 	peer.ip = ipAddress;
-	return peer;
+	return parseToJSONCompatObj(peer);
```
sameersubudhi, line 49: Seems like an issue from the SDK. For now, please update the method as below:

```
const {
	Logger,
	Signals,
	Utils: { isObject },
} = require('lisk-service-framework');

const refactorPeer = (orgPeer, state) => {
	const { chainID, ipAddress, options: { height } = {}, ...peer } = orgPeer;
	peer.chainID = isObject(chainID)
		? Buffer.from(chainID.data).toString('hex')
		: chainID;
	peer.state = state;
	peer.height = height;
	peer.ip = ipAddress;
	return peer;
};
```

sameersubudhi (changes requested):

services/blockchain-connector/shared/sdk/peerCache.js
sameersubudhi: Created issue in the SDK repo to fix the same: https://github.com/LiskHQ/lisk-sdk/issues/8173.
Please add a TODO to revert this change, once SDK addressed the above issue.

sameersubudhi (approved):
Works as expected 👍

PR #1516, opened 2023-02-22
Fix genesis block retrieval implementation in connector

### LINKED ISSUE #1512: Fix genesis block retrieval implementation in connector
### Actual behavior

- Connector is not able to start in case of using downloaded genesis block from filesystem. 
- `getGenesisBlock` method at `services/blockchain-connector/shared/sdk/genesisBlock.js` still has old implementation which needs to refactor with the SDKv6 block.

### Expected behavior

Connector should start without any error:
- Using the Genesis block downloaded into the filesystem or
- Retrieve block directly from the node

### Steps to reproduce

Start connector while using the downloaded `genesis_block.json` at `services/blockchain-connector/data`

v0.7.0-alpha.7

nagdahimanshu:
### What was the problem?
This PR resolves #1512 

### How was it solved?

- [x] Remove old implementation from `getGenesisBlock` in connector service
- [x] Connector should start without issue in case of using genesis block from file system
- [x] Add environment variable for genesis height
- [x] Code cleanup

### How was it tested?

- Local
- pm2


sameersubudhi (approved):
LGTM 👍 
The indexing however fails against the mainnet genesis block. This will be handled as part of https://github.com/LiskHQ/lisk-service/issues/1517.

PR #1530, opened 2023-03-13
Events endpoint does not return all events

### LINKED ISSUE #1528: Events endpoint does not return all events
### Actual behavior

`/events` endpoint returns only limited number of events (only 1 event when the latest block contains no transactions).

### Expected behavior

The endpoints should return all the events emitted by Lisk Core, when queried with pagination.

### Steps to reproduce

- Run Lisk Service against Lisk Core v4.0.0-alpha.15
- Invoke the above endpoint after the finalized height starts to increase (approx. after block height 135)

Lisk Service v0.7.0-alpha.9

nagdahimanshu:
### What was the problem?
This PR resolves #1528 

### How was it solved?

- [x] The endpoints returns all the events emitted by Lisk Core, when queried with pagination:
  - [x]  Update `deleteEventsTillHeight` method to delete only serialised events until finalised height
  - [x] Add job to schedule `deleteEventsTillHeight` every 5 minutes
- [x] Update MySQL  implementation
  - [x] Implement update method
  - [x] Update `mapRowsBySchema` method
  - [x] Add test cases
  - [x] Publish framework to GitHub
- [x] Update all dependencies
- [x] Add config for devnet mainchain URL
- [x] Update test schema for transaction dry run

### How was it tested?
Local + pm2


framework/test/functional/mysql.test.js
```diff
@@ -203,6 +203,23 @@ describe('Test MySQL', () => {
 			const distinctResult = await testTable.find({ distinct: 'id' }, 'id');
 			expect(result.length).toBeGreaterThan(distinctResult.length);
 		});
+
+		it('Update method', async () => {
```
sameersubudhi, line 207: Please update the tests as dicussed offline.

services/blockchain-indexer/shared/utils/eventsUtils.js
```diff
@@ -76,21 +74,29 @@ const getEventsInfoToIndex = async (block, events) => {
 	return eventsInfoToIndex;
 };
 
-const deleteEventsTillHeight = async (blockHeight, dbTrx) => {
+const deleteEventsTillFinalizedHeight = async () => {
```
sameersubudhi, line 77:
```suggestion
const deleteEventStrTillFinalizedHeight = async () => {
```

Also, this entire logic should be part of an explicit MySQL transaction, including the KV store updates.

services/blockchain-indexer/jobs/dataService/events.js
```diff
@@ -0,0 +1,37 @@
...
+module.exports = [
+	{
+		name: 'delete.serializedEvents',
+		description: 'Delete the serialized events until finalized height.',
+		schedule: '*/5 * * * *', // Every 5 min
+		controller: async () => {
+			try {
+				if (!config.db.isPersistEvents) {
+					logger.info('Deleting the serialized events until finalized height...');
+					await deleteEventsTillFinalizedHeight();
```
sameersubudhi, line 30:
```suggestion
					logger.debug('Deleting the serialized events until finalized height...');
					await deleteEventsTillFinalizedHeight();
					logger.info('Deleted the serialized events until finalized height...');
```

sameersubudhi (changes requested):

framework/src/mysql.js
```diff
@@ -116,7 +117,8 @@ const mapRowsBySchema = async (rawRows, schema) => {
 		const row = {};
 		Object.keys(schema).forEach(o => {
 			const val = item[o];
-			if (val || val === 0 || val === false) row[o] = getValue(cast(val, schema[o].type));
+			const valType = schema[o].type;
+			if (`${o}` in item) row[o] = getValue(cast(val, valType));
```
priojeetpriyom, line 121: maybe we can cast inside the `getValue` method itself. so the call will be more neat here. ex: `getValue(val, valType)`. Also, renaming `o` to `key` would be nice.

framework/src/mysql.js
```diff
@@ -368,6 +370,27 @@ const getTableInstance = async (tableName, tableConfig, connEndpoint = CONN_ENDP
...
 
+	const update = async (params, trx) => {
+		let isDefaultTrx = false;
+		if (!trx) {
+			trx = await createDefaultTransaction(knex);
+			isDefaultTrx = true;
+		}
+
+		const { where, updates } = params;
+		const query = queryBuilder({ ...where }, tableConfig.primaryKey, trx).update({ ...updates });
```
priojeetpriyom, line 381: would it be more capsulated if we move the `update({ ...updates })` part inside the `queryBuilder` method?

services/blockchain-indexer/shared/utils/eventsUtils.js
```diff
@@ -76,21 +74,29 @@ const getEventsInfoToIndex = async (block, events) => {
...
-			to: blockHeight,
+			from: fromHeight ? fromHeight + 1 : await getGenesisHeight(),
+			to: toHeight,
 		}],
-		limit: 10000,
 	};
-	await eventTopicsTable.delete(queryParams, dbTrx);
-	await eventsTable.delete(queryParams, dbTrx);
+
+	await eventsTable.update({ where: queryParams, updates: { eventStr: null } });
```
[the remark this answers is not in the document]
sameersubudhi: Our upsert implementation doesn't currently support conditional updates and requires an additional find first. Let's create a new issue to address this later.

framework/src/mysql.js
sameersubudhi: Let's keep it as is. queryBuilder is supposed to be used for all the methods and it's better to keep only the DQL methods within it. Let's avoid adding DMLs within the queryBuilder.

sameersubudhi (approved):
LGTM 👍

priojeetpriyom (approved):

PR #1553, opened 2023-03-24
Adds default value to pos_getStakes and interoperability_getChainAcco…

### LINKED ISSUE #1535: Stakes endpoint returns a HTTP 500 error
### Actual behavior
Calling the stakes endpoint for a validator (for example, http://devnet-service.liskdev.net:9901/api/v3/pos/stakes?address=lskzr72jsfyvrj66krxg7tjmzhs28ko6okc9ufubj) returns a HTTP 500 error with the message `Server error: Key 7160f8680000049d8e6fd114b8a55241fba7c283bbc3e8c53283 does not exist.`

### Expected behavior
Calling the stakes endpoint for a validator should return the stakes details or empty data if there's none

### Steps to reproduce
- Using a validator's address as an address param, call the stakes endpoint

All

### Additional Information
The default response for `pos_getStakes` must return:
```jsonc
{
        "stakes": [],
        "pendingUnlocks": [],
}
```

And, for `interoperability_getChainAccount` must return:
```jsonc
{
        "lastCertificate": {},
        "name": undefined,
        "status": undefined,
}
```

vardan10:
…unt SDK error cases #1535

### What was the problem?
This PR resolves #1535

### How was it solved?
Added a validation on service

### How was it tested?
Stakes was tested via API
chainAccount was tested via molecular

### DIFF
--- a/services/blockchain-connector/shared/sdk/constants/defaultResponses.js
+++ b/services/blockchain-connector/shared/sdk/constants/defaultResponses.js
@@ -1,0 +1,15 @@
+const STARTER_DEFAULT_RES = {
+    stakes: [],
+    pendingUnlocks: [],
+};
+
+const CHAIN_ACCOUNT_DEFAULT_RES = {
+    lastCertificate: {},
+    name: null,
+    status: null,
+};
+
+module.exports = {
+    STARTER_DEFAULT_RES,
+    CHAIN_ACCOUNT_DEFAULT_RES
+}
--- a/services/blockchain-connector/shared/utils/regex.js
+++ b/services/blockchain-connector/shared/utils/regex.js
@@ -16,9 +16,14 @@
 const BLS_KEY = /^\b[a-fA-F0-9]{96}\b$/;
 const CHAIN_ID = /^\b[a-fA-F0-9]{8}\b$/;
 const PROOF_OF_POSSESSION = /^\b[a-fA-F0-9]{192}\b$/;
+const CHAIN_ACCOUNT_KEY_NOT_FOUND_ERROR = /^Key [a-f0-9]{20} does not exist\.$/;
+const STARTER_KEY_NOT_FOUND_ERROR = /^Key [a-f0-9]{52} does not exist\.$/;
+
 
 module.exports = {
 	BLS_KEY,
 	CHAIN_ID,
 	PROOF_OF_POSSESSION,
+	CHAIN_ACCOUNT_KEY_NOT_FOUND_ERROR,
+	STARTER_KEY_NOT_FOUND_ERROR
 };
--- a/services/blockchain-connector/shared/sdk/interoperability.js
+++ b/services/blockchain-connector/shared/sdk/interoperability.js
@@ -19,12 +19,19 @@ const {
 } = require('lisk-service-framework');
 
 const { timeoutMessage, invokeEndpoint } = require('./client');
+const regex = require('../utils/regex');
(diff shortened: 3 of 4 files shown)


services/blockchain-connector/shared/sdk/constants/defaultResponses.js
```diff
@@ -0,0 +1,15 @@
+const STARTER_DEFAULT_RES = {
```
sameersubudhi, line 1:
```suggestion
const DEFAULT_RES_GET_STAKER = {
```

services/blockchain-connector/shared/sdk/constants/defaultResponses.js
```diff
@@ -0,0 +1,15 @@
...
+	stakes: [],
+	pendingUnlocks: [],
+};
+
+const CHAIN_ACCOUNT_DEFAULT_RES = {
```
sameersubudhi, line 6:
```suggestion
const DEFAULT_RES_GET_CHAIN_ACCOUNT = {
```

services/blockchain-connector/shared/utils/regex.js
```diff
@@ -16,9 +16,13 @@
 const BLS_KEY = /^\b[a-fA-F0-9]{96}\b$/;
 const CHAIN_ID = /^\b[a-fA-F0-9]{8}\b$/;
 const PROOF_OF_POSSESSION = /^\b[a-fA-F0-9]{192}\b$/;
+const CHAIN_ACCOUNT_KEY_NOT_FOUND_ERROR = /^Key [a-f0-9]{20} does not exist\.$/;
+const STARTER_KEY_NOT_FOUND_ERROR = /^Key [a-f0-9]{52} does not exist\.$/;
```
sameersubudhi, line 20: Please use a single regex.
```suggestion
const KEY_NOT_EXIST = /^Key [a-f0-9]+ does not exist\.$/;
```

sameersubudhi (changes requested):

services/blockchain-connector/shared/sdk/constants/defaultResponses.js
```diff
@@ -0,0 +1,35 @@
...
+const DEFAULT_RES_GET_STAKER = {
+	stakes: [],
+	pendingUnlocks: [],
+};
+
+const DEFAULT_RES_GET_CHAIN_ACCOUNT = {
+	lastCertificate: {},
+	name: null,
+	status: null,
+};
```
priojeetpriyom, line 26: Will these be reused somewhere? If not, then maybe we can define these in the respective files? Also, do we need a constant for the default structure? Usually, we just hardcode it in other cases. Ex: https://github.com/LiskHQ/lisk-service/blob/feature/1157-prepare-lisk-service-v070-for-beta/services/blockchain-indexer/shared/dataService/business/pos/stakers.js#L43-L51

vardan10: Currently not reused anywhere. Will apply the changes.

sameersubudhi: Leaving it as is would've been fine because the fix from our end was most likely temporary. The changes will be rolled back once the SDK fixes it at their end. But let's leave it as is.

sameersubudhi (approved):
LGTM and works as expected 👍

priojeetpriyom (approved):

PR #1561, opened 2023-03-31
Add job in blockchain-app-registry to cleanup empty folders inside data directory

### LINKED ISSUE #1537: Add job in blockchain-app-registry to cleanup empty folders inside data directory
### Description

When an app is deleted from github app-registry repository, the files are delete from locally in a current sync job. But the empty folders are left behind. Add a new job which will loop through all folders inside data directory, delete any non-metadata related files and then delete empty folders. Discuss with team about frequency of the job.

### Acceptance Criteria
- A cron job is defined in blockchain-app-registry and working fine
- Non metadata files are deleted. (i.e. .gitkeep)
- Empty folders inside data directory are deleted

vardan10:
### What was the problem?
This PR resolves #1537

### How was it solved?

- [x]  A cron job is defined in blockchain-app-registry.
- [x]  Non metadata files (i.e. .gitkeep) are deleted by the cron job.
- [x]  Empty folders inside data directory are deleted by the cron job.

### How was it tested?

- [x] Local


services/blockchain-app-registry/jobs/clean.js
```diff
@@ -0,0 +1,37 @@
...
+const logger = require('lisk-service-framework').Logger();
+const path = require('path');
+
+const { deleteEmptyFoldersAndNonMetaFiles } = require('../shared/utils/fsUtils');
+
+module.exports = [
+	{
+		name: 'job.clean.data',
+		description: 'Delete any non-metadata files and empty folders inside data directory',
+		schedule: '0 0 * * *', // Every day at midnight
```
vardan10, line 25: This job runs every midnight, which should be fine according to me to cleanup data directory. Opinions are welcome.

services/blockchain-app-registry/shared/constants.js
```diff
@@ -19,7 +19,10 @@ const KV_STORE_KEY = {
 	COMMIT_HASH_UNTIL_LAST_SYNC: 'commitHashUntilLastSync',
 };
 
+const NON_META_FILES = ['.gitKeep'];
```
vardan10, line 22: Do we have any other files except .gitKeep to delete?

services/blockchain-app-registry/jobs/clean.js
```diff
@@ -0,0 +1,37 @@
+/*
+ * LiskHQ/lisk-service
+ * Copyright © 2022 Lisk Foundation
```
nagdahimanshu, line 3:
```suggestion
 * Copyright © 2023 Lisk Foundation
```

services/blockchain-app-registry/shared/constants.js
sameersubudhi: Please do the opposite instead! Kindly make a list of files that should be kept instead and the others not in the list can be removed.

services/blockchain-app-registry/shared/utils/fsUtils.js
```diff
@@ -179,6 +179,25 @@ const rename = async (oldName, newName) => new Promise((resolve, reject) => {
 	});
 });
 
+const deleteEmptyFoldersAndNonMetaFiles = (directoryPath) => {
```
sameersubudhi, line 182: Could we please this method asynchronous. Please use the async versions for the following methods: `readdirSync`, `statSync`, `readdirSync` and `unlinkSync`. You can refer to the existing implementations for the same in other microservices. If they do not exist, please implement the same in `fsUtils.js`. Please make sure to add tests for any methods you add to fsUtils.

services/blockchain-app-registry/tests/unit/shared/utils/fsUtils.test.js
```diff
@@ -94,6 +95,23 @@ describe('Test filesystem util methods', () => {
...
+		await mkdir(subDirPath);
+		expect(exists(subDirPath)).resolves.toBe(true);
+
+		const subFilePath = path.join(dirPath, '.gitKeep');
+
+		expect(exists(subFilePath)).resolves.toBe(false);
+		await write(subFilePath, JSON.stringify(testData));
+		expect(exists(subFilePath)).resolves.toBe(true);
+
+		deleteEmptyFoldersAndNonMetaFiles(dirPath);
```
sameersubudhi, line 111:
```suggestion
		const subFilePath = path.join(dirPath, '.gitKeep');
		expect(exists(subFilePath)).resolves.toBe(false);
		await write(subFilePath, JSON.stringify(testData));
		expect(exists(subFilePath)).resolves.toBe(true);

		await deleteEmptyFoldersAndNonMetaFiles(dirPath);
```

services/blockchain-app-registry/shared/constants.js
```diff
@@ -19,7 +19,10 @@ const KV_STORE_KEY = {
 	COMMIT_HASH_UNTIL_LAST_SYNC: 'commitHashUntilLastSync',
 };
 
+const META_FILES = ['app.json', 'nativetokens.json', '.png', '.svg'];
```
sameersubudhi, line 22: Maybe split them and rename the resulting constants to `ALLOWED_FILES` and `ALLOWED_FILE_EXTENSIONS`.

services/blockchain-app-registry/jobs/clean.js
```diff
@@ -0,0 +1,37 @@
...
+const logger = require('lisk-service-framework').Logger();
+const path = require('path');
+
+const { deleteEmptyFoldersAndNonMetaFiles } = require('../shared/utils/fsUtils');
+
+module.exports = [
+	{
+		name: 'job.clean.data',
+		description: 'Delete any non-metadata files and empty folders inside data directory',
+		schedule: '0 0 * * *', // Every day at midnight
```
sameersubudhi, line 25:
```suggestion
		name: 'delete.non.metadata.files',
		description: 'Delete any non-metadata files and empty folders inside data directory.',
		schedule: '0 0 * * *', // Every day at midnight
```

sameersubudhi (changes requested):

services/blockchain-app-registry/shared/utils/fsUtils.js
```diff
@@ -179,6 +179,40 @@ const rename = async (oldName, newName) => new Promise((resolve, reject) => {
...
+		const isDirectory = (await fs.promises.lstat(filePath)).isDirectory();
+
+		if (isDirectory) {
+			await deleteEmptyFoldersAndNonMetaFiles(filePath);
+			await deleteFolderIfEmpty(filePath);
+		} else {
+			const fileName = path.basename(filePath);
+
+			if (!ALLOWED_FILE_EXTENSIONS.some((ending) => file.endsWith(ending))
+					&& !ALLOWED_FILES.some((name) => fileName === name)) {
```
priojeetpriyom, line 207: Can we reuse this method to avoid duplicate logic. Currently we use this method while syncing with new commit in remote repo to determine if we should download the added/updated file. So, currently we are not downloading the added/updated logo image. Could we please update the logic so that we consider both meta files and extensions?
https://github.com/LiskHQ/lisk-service/blob/feature/1157-prepare-lisk-service-v070-for-beta/services/blockchain-app-registry/shared/utils/downloadRepository.js#L201-L203

services/blockchain-app-registry/shared/utils/fsUtils.js
```diff
@@ -179,6 +179,40 @@ const rename = async (oldName, newName) => new Promise((resolve, reject) => {
...
+				await fs.promises.unlink(filePath);
```
priojeetpriyom, line 208: there is already `rm` defined in this file. maybe we can use that one?

services/blockchain-app-registry/shared/utils/fsUtils.js
priojeetpriyom, line 1: Would it be more neat if we define a function for `fs` endpoints instead of invoking these directly?

services/blockchain-app-registry/shared/constants.js
```diff
@@ -19,7 +19,12 @@ const KV_STORE_KEY = {
 	COMMIT_HASH_UNTIL_LAST_SYNC: 'commitHashUntilLastSync',
 };
 
+const ALLOWED_FILES = ['app.json', 'nativetokens.json'];
+const ALLOWED_FILE_EXTENSIONS = ['.png', '.svg'];
```
priojeetpriyom, line 23: 1. Maybe we can reuse the allowed files defined in config?
https://github.com/LiskHQ/lisk-service/blob/feature/1157-prepare-lisk-service-v070-for-beta/services/blockchain-app-registry/config.js#L81-L84

2. Also, maybe we can move the both list in one file (maybe config?) and then only export one variable as the concatenation of these two? Ex: `ALLOWED_META_FILES_AND_EXTENSION` which is a concatenation of `ALLOWED_FILES` and `ALLOWED_FILE_EXTENSIONS`

services/blockchain-app-registry/shared/utils/fsUtils.js
```diff
@@ -179,6 +179,40 @@ const rename = async (oldName, newName) => new Promise((resolve, reject) => {
 	});
 });
 
+const deleteFolderIfEmpty = async (folderPath) => {
+	const files = await fs.promises.readdir(folderPath);
```
nagdahimanshu, line 183: We already have method implemented to get all files. We can re-use that method here:

```suggestion
	const files = await getFiles(folderPath);
```

services/blockchain-app-registry/shared/utils/fsUtils.js
```diff
@@ -179,6 +179,40 @@ const rename = async (oldName, newName) => new Promise((resolve, reject) => {
...
+
+	if (files.length === 0) {
+		await fs.promises.rmdir(folderPath);
+		logger.trace(`Deleted folder: ${folderPath}`);
```
nagdahimanshu, line 187:
```suggestion
		logger.trace(`Deleted folder: ${folderPath}.`);
```

services/blockchain-app-registry/shared/utils/fsUtils.js
```diff
@@ -179,6 +179,40 @@ const rename = async (oldName, newName) => new Promise((resolve, reject) => {
...
+		if (isDirectory) {
+			await deleteEmptyFoldersAndNonMetaFiles(filePath);
+			await deleteFolderIfEmpty(filePath);
+		} else {
+			const fileName = path.basename(filePath);
+
+			if (!ALLOWED_FILE_EXTENSIONS.some((ending) => file.endsWith(ending))
+					&& !ALLOWED_FILES.some((name) => fileName === name)) {
+				await fs.promises.unlink(filePath);
+				logger.trace(`Deleted file: ${filePath}`);
```
nagdahimanshu, line 209:
```suggestion
				logger.trace(`Deleted file: ${filePath}.`);
```

services/blockchain-app-registry/shared/constants.js
sameersubudhi: > Maybe we can reuse the allowed files defined in config?
Agreed!

> Also, maybe we can move the both list in one file (maybe config?) and then only export one variable as the concatenation of these two? Ex: ALLOWED_META_FILES_AND_EXTENSION which is a concatenation of ALLOWED_FILES and ALLOWED_FILE_EXTENSIONS

Please move the constants to the config file as advised but keep ALLOWED_FILES and ALLOWED_FILE_EXTENSIONS separate since they contain different types of data.

ALLOWED_FILES can be defined as `Object.values(config.FILENAME)`

services/blockchain-app-registry/shared/utils/fsUtils.js
sameersubudhi, line 1: Thanks for pointing it out @priojeetpriyom.

@vardan10 fsUtils must contain methods such as `rm`, `rmdir`, `ls`, `mkdir` etc. Everything specific to the application must be handled outside of here.

services/blockchain-app-registry/shared/utils/fsUtils.js
vardan10: Confirmed with @sameersubudhi, not to modify code from download repository. Maybe better to create a separate issue for this.

services/blockchain-app-registry/jobs/deleteNonMetadata.js
```diff
@@ -0,0 +1,66 @@
...
+const { isMetadataFile } = require('../shared/utils/downloadRepository');
+
+const deleteFolderIfEmpty = async (folderPath) => {
+	const files = await getFilesAndDirs(folderPath);
+
+	if (files.length === 0) {
+		await rmdir(folderPath);
+		logger.trace(`Deleted folder: ${folderPath}.`);
+	}
+};
```
sameersubudhi, line 28: Let us please stick to `directory`instead of `folder` for consistency.
```suggestion
const removeDirectoryIfEmpty = async (dirPath) => {
	const files = await getFilesAndDirs(dirPath);

	if (files.length === 0) {
		await rmdir(dirPath);
		logger.trace(`Removed directory: ${dirPath}.`);
	}
};
```

services/blockchain-app-registry/jobs/deleteNonMetadata.js
```diff
@@ -0,0 +1,66 @@
...
+			await deleteEmptyFoldersAndNonMetaFiles(filePath);
+			await deleteFolderIfEmpty(filePath);
+		} else if (!ALLOWED_FILE_EXTENSIONS.some((ending) => filePath.endsWith(ending))
+					&& !isMetadataFile(filePath)) {
+			await rm(filePath);
+			logger.trace(`Deleted file: ${filePath}.`);
+		}
+		/* eslint-enable no-await-in-loop */
+	}
+};
```
sameersubudhi, line 48:
```suggestion
const removeEmptyDirectoriesAndNonMetaFiles = async (dirPath) => {
	const contents = await getFilesAndDirs(dirPath);

	for (let i = 0; i < contents.length; i++) {
		/* eslint-disable no-await-in-loop */
		const filePath = contents[i];
		const isDirectory = (await stats(filePath)).isDirectory();

		if (isDirectory) {
			await removeEmptyDirectoriesAndNonMetaFiles(filePath);
			await removeDirectoryIfEmpty(filePath);
		} else if (!ALLOWED_FILE_EXTENSIONS.some((ending) => filePath.endsWith(ending))
					&& !isMetadataFile(filePath)) {
			await rm(filePath);
			logger.trace(`Removed file: ${filePath}.`);
		}
		/* eslint-enable no-await-in-loop */
	}
};
```

services/blockchain-app-registry/jobs/deleteNonMetadata.js
```diff
@@ -0,0 +1,66 @@
...
+
+module.exports = [
+	{
+		name: 'delete.non.metadata.files',
+		description: 'Delete any non-metadata files and empty folders inside data directory',
```
sameersubudhi, line 53: Please ensure to always end a comment/description with a `.`.
```suggestion
		description: 'Delete any non-metadata files and empty folders inside data directory.',
```

sameersubudhi (changes requested):

services/blockchain-app-registry/shared/utils/fsUtils.js
sameersubudhi: The existing piece of code can be left as is and requires no change.

sameersubudhi (approved):
Code changes look fine but needs proper testing to ensure that it doesn't cause any issue with the metadata index update job.

nagdahimanshu (approved):
Looks and works as expected 👍

priojeetpriyom (approved):

PR #1570, opened 2023-04-06
Incorrect totalLocked and totalSelfStaked values

### LINKED ISSUE #1546: Incorrect totalLocked and totalSelfStaked values
### Actual behavior

```jsonc
 {
      "basic": {
        "totalPeers": 0,
        "connectedPeers": 0,
        "disconnectedPeers": 0
      },
      "height": {},
      "networkVersion": {},
      "totalLocked": [
        {
          "tokenID": "0400000000000000",
          "amount": "-18700000000" // <-- Shouldn't be negative
        }
      ],
      "totalStaked": {
        "tokenID": "0400000000000000",
        "amount": "2000000000"
      },
      "totalSelfStaked": {
        "tokenID": "0400000000000000",
        "amount": null // <-- Shoudn't be null, must be 0 instead
      }
 }
```

### Expected behavior

`totalLocked` and `totalSelfStaked` values must always be  greater than or equal to `0`. 

### Steps to reproduce

- Query `/network/statistics` endpoint

Lisk Service v0.7.0-alpha.10

priojeetpriyom:
### What was the problem?
This PR resolves #1546 

### How was it solved?
- [x] index total locked amount for genesis block
- [x] index total staked and total self staked for genesis block
- [x] Fix missing data for 3 keys in following endpoint: http://devnet-service.liskdev.net:9901/api/v3/blockchain/apps/statistics
- [x] Fix null return value for totalStaked
- [x] Fix duplicate scheduling of genesis block 

### How was it tested?
- [x] Local


services/blockchain-indexer/shared/utils/blockchainIndex.js
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+const BluebirdPromise = require('bluebird');
+
+const { KV_STORE_KEY } = require('../constants');
+const keyValueTable = require('../database/mysqlKVStore');
+
+const updateTotalLockedAmounts = (tokenIDLockedAmountChangeMap, dbTrx) => BluebirdPromise.map(
```
sameersubudhi, line 21:
```suggestion
const updateTotalLockedAmounts = async (tokenIDLockedAmountChangeMap, dbTrx) => BluebirdPromise.map(
```

services/blockchain-indexer/shared/utils/pos.js
```diff
@@ -21,7 +21,7 @@ const { KV_STORE_KEY } = require('../constants');
 
 const getStakeInfo = async (keyPrefix) => {
 	const [row] = await keyValueTable.getByPattern(keyPrefix);
-	const stakeAmount = row && row.value ? row.value.toString() : null;
+	const stakeAmount = row && row.value ? row.value.toString() : 0;
```
sameersubudhi, line 24:
```suggestion
	const stakeAmount = row && row.value ? row.value.toString() : '0';
```

services/blockchain-connector/shared/sdk/genesisBlock.js
```diff
@@ -83,9 +84,68 @@ const getGenesisConfig = async () => {
...
+				},
+			}]
+			: [];
+	}
+
+	if (params.module) {
+		return assetModule ? [assetModule] : [];
+	}
+
+	return genesisBlock.assets;
```
sameersubudhi, line 114: This will result in inconsistency in response structure. Please always return the response in the following format:
```
{
 "subStore1": [],
 "subStore2": [],
}
```

services/blockchain-connector/shared/sdk/genesisBlock.js
```diff
@@ -83,9 +84,68 @@ const getGenesisConfig = async () => {
 	}
 };
 
+const getGenesisAssets = async (params = {}) => {
+	const rawGenesisBlock = await getGenesisBlock(true);
+	const genesisBlock = await formatBlock(rawGenesisBlock);
+
+	const assetModule = genesisBlock.assets.find(
+		genesisAsset => genesisAsset.module === params.module,
+	);
```
sameersubudhi, line 93:
```suggestion
	const assetsByModule = genesisBlock.assets.find(entry => entry.module === params.module);
```

services/blockchain-connector/shared/sdk/genesisBlock.js
```diff
@@ -83,9 +84,68 @@ const getGenesisConfig = async () => {
...
+	moduleName1: {
+		module1.data.key1 (subStore): module1.data.key1.length,
+	}
+	moduleName2: {
+		module2.data.key2 (subStore): module2.data.key2.length,
+	}
+}
+*/
+const getGenesisAssetsLength = async (params) => {
+	const genesisAssets = await getGenesisAssets(params);
```
sameersubudhi, line 128: This would not return the desired response when no params are passed. Please re-verify. Also, the corresponding tests are missing.

services/blockchain-indexer/shared/indexer/genesisBlock.js
```diff
@@ -0,0 +1,112 @@
...
+const indexTokenModuleAssets = async (dbTrx) => {
+	const genesisBlockAssetsLength = await requestConnector('getGenesisAssetsLength', {
+		module: MODULE.TOKEN,
+		subStore: MODULE_SUB_STORE.TOKEN.USER,
+	});
+	const totalUsers = genesisBlockAssetsLength[MODULE.TOKEN][MODULE_SUB_STORE.TOKEN.USER];
+
+	const tokenIDLockedAmountChangeMap = {};
+	let offset = 0;
+	while (offset < totalUsers) {
```
sameersubudhi, line 43: Please use the `requestAll` utility instead of this while loop.

services/blockchain-indexer/shared/indexer/genesisBlock.js
```diff
@@ -0,0 +1,112 @@
...
+	POS: {
+		STAKERS: 'stakers',
+	},
+});
+
+const indexTokenModuleAssets = async (dbTrx) => {
+	const genesisBlockAssetsLength = await requestConnector('getGenesisAssetsLength', {
+		module: MODULE.TOKEN,
+		subStore: MODULE_SUB_STORE.TOKEN.USER,
+	});
```
sameersubudhi, line 38:
```suggestion
	const genesisBlockAssetsLength = await requestConnector(
		'getGenesisAssetsLength',
		 { module: MODULE.TOKEN,	subStore: MODULE_SUB_STORE.TOKEN.USER	},
	);
```

services/blockchain-indexer/shared/indexer/genesisBlock.js
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		offset += userSubStoreInfos.length;
+	}
+
+	await updateTotalLockedAmounts(tokenIDLockedAmountChangeMap, dbTrx);
+};
+
+const indexPosModuleAssets = async (dbTrx) => {
+	const genesisBlockAssetsLength = await requestConnector('getGenesisAssetsLength', { module: MODULE.POS, subStore: MODULE_SUB_STORE.POS.STAKERS });
```
sameersubudhi, line 72:
```suggestion
	const genesisBlockAssetsLength = await requestConnector(
		'getGenesisAssetsLength',
		{ module: MODULE.POS, subStore: MODULE_SUB_STORE.POS.STAKERS },
	);
```

services/blockchain-indexer/shared/indexer/genesisBlock.js
```diff
@@ -0,0 +1,112 @@
...
+	const totalStakers = genesisBlockAssetsLength[MODULE.POS][MODULE_SUB_STORE.POS.STAKERS];
+
+	let offset = 0;
+	let totalStakeChange = BigInt(0);
+	let totalSelfStakeChange = BigInt(0);
+
+	while (offset < totalStakers) {
```
sameersubudhi, line 79: Same as above.

sameersubudhi (changes requested):

services/blockchain-indexer/shared/indexer/genesisBlock.js
priojeetpriyom: Applied.

services/blockchain-connector/shared/sdk/genesisBlock.js
priojeetpriyom: Seems to respect the response structure. This the response when no empty params is passed. Will add test coverage.

```
{
    "interoperability":{
        "outboxRootSubstore":0,
        "chainDataSubstore":0,
        "channelDataSubstore":0,
        "chainValidatorsSubstore":0,
        "ownChainDataSubstore":1,
        "terminatedStateSubstore":0,
        "terminatedOutboxSubstore":0,
        "registeredNamesSubstore":1
    },
    "legacy":{
        "accounts":30
    },
    "pos":{
        "validators":103,
        "stakers":103
    },
    "token":{
        "userSubstore":104,
        "supplySubstore":1,
        "escrowSubstore":0,
        "supportedTokensSubstore":0
    }
}
```

services/blockchain-connector/shared/sdk/genesisBlock.js
priojeetpriyom: As per previous discussion this method was supposed to return an array of asset filtered by module and subStore key. seems to be returning that structure for empty param, module, module and subStore filtering. Will rename the endpoint to `getAssetByModule` and return the suggested structure.

services/blockchain-indexer/shared/indexer/genesisBlock.js
priojeetpriyom: Applied.

services/blockchain-connector/methods/blocks.js
```diff
@@ -83,4 +85,28 @@ module.exports = [
...
+		name: 'getGenesisAssetByModule',
+		controller: async ({ module, subStore, limit, offset }) => getGenesisAssetByModule({
+			module,
+			subStore,
+			limit,
+			offset,
+		},
+		),
+		params: {
+			module: { type: 'string' },
```
nagdahimanshu, line 98:
```suggestion
			module: { type: 'string', optional: true },
```

services/blockchain-connector/shared/sdk/genesisBlock.js
```diff
@@ -83,9 +90,81 @@ const getGenesisConfig = async () => {
...
+			moduleData = moduleData.slice(params.offset, params.offset + params.limit);
+		}
+
+		return [{
+			...assetByModule,
+			data: {
+				[params.subStore]: moduleData,
+			},
+		}];
+	}
```
nagdahimanshu, line 119: May be we can update this and avoid using `let` if possible. Also, `limit` and `offset` will be always available as we are setting default values in controller
```suggestion
		if (params.subStore) {
		// Return empty array otherwise if passed subStore is not present
		const moduleData = assetByModule.data[params.subStore]
			? assetByModule.data[params.subStore]
			: [];

		return [{
			...assetByModule,
			data: {
				[params.subStore]: moduleData.slice(params.offset, params.offset + params.limit),
			},
		}];
	}
```

services/blockchain-connector/shared/sdk/genesisBlock.js
```diff
@@ -83,9 +90,81 @@ const getGenesisConfig = async () => {
...
+}
+*/
+const getGenesisAssetByModule = async (params = {}) => {
+	const genesisAssets = await getGenesisAssets(params);
+
+	if (genesisAssets && genesisAssets.length) {
+		return genesisAssets[0].data;
+	}
+	return {};
+};
```
nagdahimanshu, line 138: May be we can be update this:
```suggestion
const getGenesisAssetByModule = async (params = {}) => {
	const [genesisAssets = {}] = await getGenesisAssets(params);
	return genesisAssets.data ? genesisAssets.data : genesisAssets;
};
```

services/blockchain-connector/methods/blocks.js
priojeetpriyom: If we do not require the module, then we would fetch data for all module which could lead to the time out issue. That is why module is a required parameter here.

nagdahimanshu: In that case, can we explicitly set `optional` to false
```
module: { type: 'string', optional: false },
```

services/blockchain-connector/shared/sdk/genesisBlock.js
priojeetpriyom: limit and offset will not be present always. ex: https://github.com/LiskHQ/lisk-service/blob/1546-incorrect-total-locked-and-total-self-staked-values/services/blockchain-connector/shared/sdk/genesisBlock.js#L148

services/blockchain-connector/shared/sdk/genesisBlock.js
priojeetpriyom: Applied.

services/blockchain-connector/methods/blocks.js
priojeetpriyom: params are required by default, that is why removed the explicit declaration after adding.

services/blockchain-connector/tests/functional/genesisBlock.test.js
```diff
@@ -59,3 +60,51 @@ describe('Genesis Block import tests', () => {
 		expect(expectedHash).toEqual(fileHash);
 	});
 });
+
+describe('Test getGenesisAssets method', () => {
+	const module = 'token';
+	const subStore = 'userSubstore';
```
nagdahimanshu, line 66: Let's define those variables globally as it is required by other test cases.

services/blockchain-indexer/shared/indexer/genesisBlock.js
```diff
@@ -0,0 +1,112 @@
...
+	);
+	const totalUsers = genesisBlockAssetsLength[MODULE.TOKEN][MODULE_SUB_STORE.TOKEN.USER];
+
+	const tokenModuleData = await requestAll(
+		requestConnector,
+		'getGenesisAssetByModule',
+		{ module: MODULE.TOKEN, subStore: MODULE_SUB_STORE.TOKEN.USER },
+		totalUsers,
+	);
+	const userSubStoreInfos = tokenModuleData[MODULE_SUB_STORE.TOKEN.USER];
```
nagdahimanshu, line 48: Just an suggestion, this can be updated to:
```suggestion
	const { userSubstore: userSubStoreInfos } = await requestAll(
		requestConnector,
		'getGenesisAssetByModule',
		{ module: MODULE.TOKEN, subStore: MODULE_SUB_STORE.TOKEN.USER },
		totalUsers,
	);
```

priojeetpriyom: Thanks for the suggestion. I would prefer keeping keeping it as we are using the constant to access the property instead of hardcoding.

services/blockchain-indexer/shared/indexer/genesisBlock.js
```diff
@@ -0,0 +1,112 @@
...
+	let totalStakeChange = BigInt(0);
+	let totalSelfStakeChange = BigInt(0);
+
+	const posModuleData = await requestAll(
+		requestConnector,
+		'getGenesisAssetByModule',
+		{ module: MODULE.POS, subStore: MODULE_SUB_STORE.POS.STAKERS },
+		totalStakers,
+	);
+	const stakersInfo = posModuleData[MODULE_SUB_STORE.POS.STAKERS];
```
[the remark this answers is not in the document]
priojeetpriyom: https://github.com/LiskHQ/lisk-service/pull/1570#discussion_r1168308463

nagdahimanshu (approved):

sameersubudhi (approved):

PR #1598, opened 2023-04-26
Fix api response inconsistencies

### LINKED ISSUE #1597: Fix api response inconsistencies
### Actual behavior
/blockchain/apps/meta/tokens

This call fails. http://localhost:9901/api/v3/blockchain/apps/meta/tokens?chainName=Lisk&tokenName=lsk&limit=10&offset=0&sort=chainName%3Aasc

/pos/stake

meta.staker.publicKey is null even when queried with publicKey: http://localhost:9901/api/v3/pos/stakes?publicKey=8cd956764adebf71c62824bc1ae5cd11373e5f38016fe08fcf0d12d524e2d0ad

### Expected behavior
- The errors should be resolved

### Steps to reproduce
- Start lisk service and invoke related endpoints

- lisk-service 0.7.0

priojeetpriyom:
### What was the problem?
This PR resolves #1597

### How was it solved?
- [x] /blockchain/apps/meta/tokens

This call fails. http://localhost:9901/api/v3/blockchain/apps/meta/tokens?chainName=Lisk&tokenName=lsk&limit=10&offset=0&sort=chainName%3Aasc

- [x] /pos/stake

meta.staker.publicKey is null even when queried with publicKey: http://localhost:9901/api/v3/pos/stakes?publicKey=8cd956764adebf71c62824bc1ae5cd11373e5f38016fe08fcf0d12d524e2d0ad

- [x] Exclude data folder files from eslint

### How was it tested?
- [x] Local


services/blockchain-app-registry/shared/metadata.js
```diff
@@ -251,7 +251,7 @@ const getBlockchainAppsTokenMetadata = async (params) => {
 		// chainID or chainName must be specified with the network
```
vardan10, line 251: Do we need to change this comment?

vardan10 (approved):

services/blockchain-indexer/shared/dataService/business/pos/stakes.js
```diff
@@ -86,7 +86,7 @@ const getStakes = async params => {
 	stakesResponse.meta.staker = {
 		address: params.address,
 		name: accountInfo.name || null,
-		publicKey: accountInfo.publicKey || null,
+		publicKey: accountInfo.publicKey || params.publicKey || null,
```
sameersubudhi, line 89: Could you also please ensure to asynchronously index the publicKey? We don't need to await the indexing.

.eslintignore
```diff
@@ -4,3 +4,4 @@ test
 *.gz
 *.sql
 *.zst
+services/**/data/**
```
priojeetpriyom, line 7: eslint complains about `blockchain-app-registry/data/app-registry/validation` directory depencencies. That is why added this exclusion. It can be applied to the specific folder, but thought ideally data won't need linting anyway.

priojeetpriyom, line 7: eslint complains about `blockchain-app-registry/data/app-registry/validation` directory depencencies. That is why added this exclusion. It can be applied to the specific folder, but thought ideally data won't need linting anyway.

services/blockchain-app-registry/shared/metadata.js
```diff
@@ -248,10 +248,10 @@ const getBlockchainAppsTokenMetadata = async (params) => {
 		const { tokenName, ...remParams } = params;
 		params = remParams;
 
-		// chainID or chainName must be specified with the network
+		// chainName must be specified with the network
 		// Skip when tokenID is specified, network can be resolved automatically
 		if (!('tokenID' in params) && !('chainID' in params) && ((!('chainName' in params)) || (!('network' in params)))) {
-			throw new InvalidParamsException('\'tokenName\' must be specified with either \'chainID\', or \'chainName\' and \'network\'.');
+			throw new InvalidParamsException('\'tokenName\' must be specified with either \'chainID\', or \'tokenID\', or (\'chainName\' and \'network\').');
```
nagdahimanshu, line 254: I think we can completely remove this condition. We added this condition initially when the network was supposed to be mandatory parameter which is not the case now. It works fine after removing the given condition as well. What do you say @sameersubudhi?

sameersubudhi: Yes, let's remove this check. The end-users can filter the results as per requirements. Also, please ensure that the docs are up-to-date in this regard.

priojeetpriyom: Applied.

nagdahimanshu (approved):
LGTM 👍

sameersubudhi (approved):

PR #1696, opened 2023-06-13
Extend the top balances endpoint

### LINKED ISSUE #1695: Extend the top balances endpoint
### Description

Extend the `/token/balances/top` endpoint to support the `search` query param.

### Acceptance Criteria

- `search` param support is added, similar to `/pos/stakers` endpoint
  - Must support filtering by `address`, `publicKey` or `name`
- Includes `publicKey` and `name` in the response
- Integration tests are up-to-date
- Swagger is up-to-date
- API documentation is up-to-date

### Additional Information
```jsonc
{
  "data": {
    "0000000000000000": [
      {
        "address": "lskdwsyfmcko6mcd357446yatromr9vzgu7eb8y99",
        "publicKey": "b1d6bc6c7edd0673f5fed0681b73de6eb70539c21278b300f07ade277e1962cd",  // <-- should be available
        "name": "genesis_84",  // <-- should be available
        "balance": "10000000",
        "knowledge": {
          "owner": "Genesis Account",
          "description": "Initial supply"
        }
      }
    ]
  },
  "meta": {
    "count": 10,
    "offset": 10,
    "total": 100
  }
}
```

vardan10:
### What was the problem?
This PR resolves #1695 

### How was it solved?
- [x] Added outer join query to Mysql in framework and added unit tests
- [x] Used outer join query to join tables and search by name, public key and address
- [x] Includes publicKey and name in the response
- [x] Integration tests are up-to-date
- [x] Swagger is up-to-date
- [x] API documentation is up-to-date
- [x] Added unit test in indexer for token top balances
- [x] Fixed topic filtering in events API

### How was it tested?
- [x] Local


framework/src/mysql/mysql.js
```diff
@@ -97,7 +97,7 @@ const resolveQueryParams = params => {
 	const KNOWN_QUERY_PARAMS = [
 		'sort', 'limit', 'offset', 'propBetweens', 'andWhere', 'orWhere', 'orWhereWith',
 		'whereIn', 'orWhereIn', 'whereJsonSupersetOf', 'search', 'aggregate', 'distinct',
-		'order', 'orSearch', 'count', 'whereNull', 'whereNotNull',
+		'order', 'orSearch', 'whereNull', 'whereNotNull', 'leftOuterJoin',
```
sameersubudhi, line 100: Could we please add the support for the remaining joins as well?

services/blockchain-indexer/shared/dataService/business/token.js
```diff
@@ -86,30 +84,52 @@ const getTokenTopBalances = async (params) => {
...
-			};
-		},
-		{ concurrency: tokenInfos.length },
-	);
+
+	remParams.leftOuterJoin = {
+		targetTable: 'accounts',
+		joinColumnLeft: 'account_balances.address',
+		joinColumnRight: 'accounts.address',
+	};
```
priojeetpriyom, line 96: we are not mentioning table name for input params (tokenID). could we get error for cases where both table has the same filtering column?

framework/src/mysql/mysql.js
```diff
@@ -321,6 +322,31 @@ const getTableInstance = async (...tableParams) => {
...
+
+		if (params.crossJoin) {
+			const { targetTable, joinColumnLeft, joinColumnRight } = params.crossJoin;
+
+			if (joinColumnLeft && joinColumnRight) {
+				query.crossJoin(targetTable, joinColumnLeft, joinColumnRight);
+			} else {
+				query.crossJoin(targetTable);
+			}
+		}
```
priojeetpriyom, line 348: As cross join is only supported by mysql, should we add it?
https://knexjs.org/guide/query-builder.html#crossjoin

services/blockchain-indexer/shared/dataService/business/token.js
```diff
@@ -86,30 +84,52 @@ const getTokenTopBalances = async (params) => {
 		data: {},
 		meta: {},
 	};
+
+	const { search, ...remParams } = params;
```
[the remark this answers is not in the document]
vardan10: We need to remove search from params or else it will be passed on to the database [here](https://github.com/LiskHQ/lisk-service/blob/1695-extend-the-top-balances-endpoint/services/blockchain-indexer/shared/dataService/business/token.js#L111) which will provide inconsistent results.

services/blockchain-indexer/shared/dataService/business/token.js
vardan10: Yes, if both tables have common column then we should query with `tableName.columnName`. But since `tokenID` is not present in accounts table we can just query with `tokenID`

framework/src/mysql/mysql.js
vardan10: Should we remove cross join in that case?

sameersubudhi (approved):
Looks good.

nagdahimanshu (approved):
LGTM and works as expected 👍

PR #1726, opened 2023-07-03
Unhandled exception when filename for transaction history download is a directory

### LINKED ISSUE #1716: Unhandled exception when filename for transaction history download is a directory
### Actual behavior

The /export/download endpoint, provided by the Export service, allows callers to download the transaction history file referred to by the filename parameter. When filename is a directory (e.g., "."), the file read operation in the downloadTransactionHistory method throws an EISDIR (“illegal operation on a directory, read”) error.

### Expected behavior

`staticFiles.exists` should properly validate the inputs in `services/export/shared/csvExport.js#297–306`.

### Steps to reproduce

- `curl 'localhost:9901/api/v3/export/download?filename=.'`

v0.7.0-beta.1

### Additional Information

- Handle errors related to file reads within the downloadTransactionHistory method.
- Add a check to the httpHandler method to ensure that the error corresponds to a valid HTTP status code; if it does not, pass a generic code to sendError, such as 500 Internal Server Error in `services/gateway/shared/moleculer-web/methods.js`
- Add necessary pattern checks for query params for the export API

vardan10:
### What was the problem?
This PR resolves #1716 

### How was it solved?
- [x] Handle errors related to file reads within the downloadTransactionHistory method.
- Add a check to the httpHandler method to ensure that the error corresponds to a valid HTTP status code
  (Skipped since we are already sending 500 status code if error does not corresponds to a valid HTTP status)
- [x] Added check in gateway to throw 400 incase of validation exception
- [x] Add necessary pattern checks for query params for the export API
- [x] Add unit tests

### How was it tested?
- [x] Local


services/gateway/apis/http-exports/methods/export.js
```diff
@@ -16,12 +16,13 @@
...
 const envelope = require('../../../sources/version3/mappings/stdEnvelope');
 const { transformParams, response, getSwaggerDescription } = require('../../../shared/utils');
+const regex = require('../../../shared/regex');
 
 module.exports = {
 	version: '2.0',
 	swaggerApiPath: '/export/download',
 	params: {
-		filename: { optional: false, type: 'string' },
+		filename: { optional: false, type: 'string', pattern: regex.FILEPATH },
```
sameersubudhi, line 25:
```suggestion
		filename: { optional: false, type: 'string', pattern: regex.CSV_EXPORT_FILENAME },
```

services/export/shared/csvExport.js
```diff
@@ -303,6 +303,9 @@ const downloadTransactionHistory = async ({ filename }) => {
 	const isFileExists = await staticFiles.exists(filename);
```
sameersubudhi, line 303: Please add a regex test (`CSV_EXPORT_FILENAME.test(filename)`) at the beginning to validate the `filename`.

services/export/shared/csvExport.js
```diff
@@ -303,6 +303,9 @@ const downloadTransactionHistory = async ({ filename }) => {
...
 	if (!isFileExists) throw new NotFoundException(`File ${filename} not found.`);
 
+	const isFile = await staticFiles.isFile(filename);
+	if (!isFile) throw new ValidationException(`${filename} is not a file.`);
```
sameersubudhi, line 307: Could we also ensure that we return a `400 BAD REQUEST` response from the controller when we throw the ValidationException?

sameersubudhi (changes requested):

services/export/shared/helpers/file.js
```diff
@@ -129,6 +129,8 @@ const purge = (dirPath, days) => new Promise((resolve, reject) => {
 
 const exists = async filePath => !!(await fs.promises.stat(filePath).catch(() => null));
 
+const isFile = async (filePath) => fs.lstatSync(filePath).isFile();
```
nagdahimanshu, line 132: Can we update is to use `lstat` instead of `lstatSync`?

services/export/shared/helpers/s3.js
```diff
@@ -177,6 +177,25 @@ const exists = async (fileName) => new Promise((resolve) => {
...
+const isFile = async (filePath) => {
+	const fileExists = await exists(filePath);
+	if (!fileExists) {
+		return false;
+	}
+
+	const stat = await new Promise((resolve) => {
+		minioClient.statObject(AWS_S3_BUCKET_NAME, filePath, (err, stats) => {
+			if (err) {
+				resolve(0);
```
sameersubudhi, line 189: Shouldn't the return value be an object? `stat.size` below would throw an error. Alternatively, we can directly resolve the boolean value within the callback if..else ladder.
```suggestion
				resolve({ size: 0 });
```

sameersubudhi (changes requested):

services/export/shared/helpers/file.js
```diff
@@ -67,13 +67,21 @@ const write = (filePath, content) => new Promise((resolve, reject) => {
 });
 
 const read = (filePath) => new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
vardan10: Moved this change outside read function.

services/export/shared/csvExport.js
```diff
@@ -52,6 +52,7 @@ const fields = require('./csvFieldMappings');
 
 const requestAll = require('./requestAll');
 const FilesystemCache = require('./csvCache');
+const { CSV_EXPORT_FILENAME } = require('./regex');
```
sameersubudhi, line 55:
```suggestion
const regex = require('./regex');
```

services/export/shared/csvExport.js
```diff
@@ -300,8 +301,12 @@ const downloadTransactionHistory = async ({ filename }) => {
 		meta: {},
 	};
 
-	const isFileExists = await staticFiles.exists(filename);
-	if (!isFileExists) throw new NotFoundException(`File ${filename} not found.`);
+	if (!CSV_EXPORT_FILENAME.test(filename)) {
```
sameersubudhi, line 304: Just so that it is clear and intuitive when reading through the code 🙂 
```suggestion
	if (!regex.CSV_EXPORT_FILENAME.test(filename)) {
```

sameersubudhi (approved):
LGTM 👍

nagdahimanshu (approved):
Looks good 👍

PR #1784, opened 2023-08-09
Transaction minimum fee is wrongly calculated

### LINKED ISSUE #1783: Transaction minimum fee is wrongly calculated
### Actual behavior
- On validator registration, the extra command fee sees to be doubly summed against the minimum fee value. As `150000` + `1000000000` isn't equal to `2000150000`
```
{
    "data": {
        "transaction": {
            "fee": {
                "tokenID": "0400000000000000",
                "minimum": "2000150000"
            }
        }
    },
    "meta": {
        "breakdown": {
            "fee": {
                "minimum": {
                    "byteFee": "150000",
                    "additionalFees": {
                        "validatorRegistrationFee": "1000000000"
                    }
                }
            }
        }
    }
}
```

- Minimum fee value is wrongly calculated for a non-initialised account. It seems the Account initialisation fee is doubled before been added to the byte fee that results to the minimum fee. On invoking the fee estimate endpoint on a send token transaction which was suppose to send token to a non-initialised account, I get the following response as seen below;

```
{
    "data": {
        "transaction": {
            "fee": {
                "tokenID": "0400000000000000",
                "minimum": "10165000"
            }
        }
    },
    "meta": {
        "breakdown": {
            "fee": {
                "minimum": {
                    "byteFee": "165000",
                    "additionalFees": {
                        "userAccountInitializationFee": "5000000"
                    }
                }
            }
        }
    }
}
```
As seen in the response breakdown above, `userAccountInitializationFee` has a value for `5000000` while the byteFee has `165000` expectation would be that the total minimum fee would be `5165000` but `10165000` is returned instead.

### Expected behavior

- Minimum fee should be the accurate sum of the minimum byteFee and additionalFees for a non-initialised account.
- Minimum fee should be the accurate sum of the minimum byteFee and additionalFees for a Validator registration transaction.

### Steps to reproduce
- Invoke the `/transactions/estimate-fees` endpoint against a  send token transaction  object to a non-initialised account. It would be observed that the total fee is way off by double the account initialisation fee value
- Do same from step but this time against a register validator transaction. It would be observed that the total fee is way off by double the `validatorRegistrationFee` fee value

### Additional Information
- Extend the gateway params to support the following properties in the request payload (Ref: https://github.com/LiskHQ/lisk-sdk/blob/e990c678f6d43debfc9960e2a8f5f98ceaa9b87b/elements/lisk-api-client/src/types.ts#L235):
  - module (required)
  - command (required)
  - senderPublicKey (required)
  - nonce (required)
  - fee (optional)
  - params (required)
    - All the properties must be available except `messageFee` and `messageFeeTokenID` in case of `token:transferCrossChain` transactions
  - signatures (optional)
  - id (optional)
- The mocking logic takes into account only the `required` params above. The necessary optional params should always be mocked
- The tests are updated
- Swagger and documentation are updated

nagdahimanshu:
### What was the problem?
This PR resolves #1783 #1786 

### How was it solved?

- [x] Fix minFee calculation (remove addition of additionalFees)
- [x] Extend gateway params to allow all transaction params (optional)
- [x] Mock all optional properties even if available in request params
- [x] Add logic to handle chain registration transaction
- [x] Fix issue with computing delegate rank #1786 

### How was it tested?
Local


services/gateway/apis/http-version3/methods/transactionsEstimateFees.js
```diff
@@ -29,11 +29,12 @@ module.exports = {
 			optional: false,
 			type: 'object',
 			props: {
+				id: { optional: true, type: 'string', patter: regex.HASH_SHA256 },
```
vardan10, line 32: Also we need to add this to swagger and possibly API documentation, if applicable.
```suggestion
				id: { optional: true, type: 'string', pattern: regex.HASH_SHA256 },
```

services/blockchain-indexer/tests/unit/shared/dataservice/business/transactionEstimateFees.test.js
```diff
@@ -585,4 +592,85 @@ describe('Test transaction fees estimates', () => {
...
+			const txWithRequiredProps = filterOptionalProps(undefined, optionalProps);
+			expect(Object.getOwnPropertyNames(txWithRequiredProps).length).toBe(0);
+		});
+
+		it('should return object when input is null', async () => {
+			const {
+				filterOptionalProps,
+			} = require(mockedTransactionFeeEstimatesFilePath);
+
+			const txWithRequiredProps = filterOptionalProps(undefined, optionalProps);
```
vardan10, line 641:
```suggestion
			const txWithRequiredProps = filterOptionalProps(null, optionalProps);
```

services/blockchain-indexer/tests/unit/shared/dataservice/business/transactionEstimateFees.test.js
```diff
@@ -585,4 +592,85 @@ describe('Test transaction fees estimates', () => {
...
+				command: 'transferCrossChain',
+				nonce: '7',
+				senderPublicKey: '3972849f2ab66376a68671c10a00e8b8b67d880434cc65b04c6ed886dfa91c2c',
+			};
+
+			const txWithRequiredProps = filterOptionalProps(input, optionalProps);
+			expect(txWithRequiredProps).toMatchObject(expectedResponse);
+		});
+
+		it('should return object when input is undefined', async () => {
```
vardan10, line 627:
```suggestion
		it('should return an empty object when input is undefined', async () => {
```

services/blockchain-indexer/tests/unit/shared/dataservice/business/transactionEstimateFees.test.js
```diff
@@ -585,4 +592,85 @@ describe('Test transaction fees estimates', () => {
...
+			const {
+				filterOptionalProps,
+			} = require(mockedTransactionFeeEstimatesFilePath);
+
+			const txWithRequiredProps = filterOptionalProps(undefined, optionalProps);
+			expect(Object.getOwnPropertyNames(txWithRequiredProps).length).toBe(0);
+		});
+
+		it('should return object when input is null', async () => {
```
vardan10, line 636:
```suggestion
		it('should return an empty object when input is null', async () => {
```

vardan10 (approved):

sameersubudhi (approved):
LGTM 👍

services/blockchain-indexer/shared/dataService/business/transactionsEstimateFees.js
```diff
@@ -202,6 +230,12 @@ const calcAdditionalFees = async (transaction) => {
 			};
 			additionalFees.total += BigInt(posConstants.data.validatorRegistrationFee);
 		}
+	} else if (transaction.module === MODULE.INTEROPERABILITY) {
+		if ([COMMAND.REGISTER_MAINCHAIN, COMMAND.REGISTER_SIDECHAIN].includes(transaction.command)) {
+			const { fee: chainRegistrationFee } = await requestConnector('getRegistrationFee');
```
sameersubudhi, line 235: Can we please cache this value?
```suggestion
			const { fee: chainRegistrationFee } = await requestConnector('getChainRegistrationFee');
```

services/gateway/apis/http-version3/swagger/definitions/transactions.json
```diff
@@ -590,6 +598,11 @@
 						"type": "string",
 						"example": "0"
 					},
+					"fee": {
```
vardan10, line 601: We had removed fee since it was not used. But incase we are adding it back to swagger then should maybe we add to gateway as well?
https://github.com/LiskHQ/lisk-service/blob/release/0.7.0/services/gateway/apis/http-version3/methods/transactionsEstimateFees.js#L31

vardan10 (approved):

services/gateway/apis/http-version3/swagger/definitions/transactions.json
nagdahimanshu: Done!

sameersubudhi (approved):

PR #1818, opened 2023-08-23
Gateway does not return response on upstream service error

### LINKED ISSUE #1817: Gateway does not return response on upstream service error
### Actual behavior
Currently, gateway microservice does not return any response when an upstream service throws an error.

### Expected behavior
Gateway should return error code `500` along with the respective error message.

### Steps to reproduce
 - Throw an error from an upstream microservice of gateway like indexer

- Lisk service v0.7.0-beta.4

priojeetpriyom:
### What was the problem?
This PR resolves #1817

### How was it solved?
- [x] Refactor gateway error handling to return proper response for different types of errors.

### How was it tested?
- [x] Local

### DIFF
--- a/services/gateway/app.js
+++ b/services/gateway/app.js
@@ -14,7 +14,6 @@
  *
  */
 const path = require('path');
-const { MoleculerError, MoleculerServerError } = require('moleculer').Errors;
 const {
 	Microservice,
 	Logger,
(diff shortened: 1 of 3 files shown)


vardan10 (approved):
LGTM :+1: and works as expected.

sameersubudhi (approved):

PR #1839, opened 2023-09-07
API client not subscribing to events

### LINKED ISSUE #1820: API client not subscribing to events
### Actual behavior

The WS API client doesn't emit engine events (`chain_*` type) after running for a long duration (approx 3-4 days).

### Expected behavior

Lisk Service should automatically detect such scenario and re-instantiate the API client. The behaviour should be applicable for both IPC and WS connections.

### Steps to reproduce

- Let Lisk Service on docker run for 3-4 days against the devnet, use the WS API client

Lisk Service v0.7.0-beta.4

nagdahimanshu:
### What was the problem?
This PR resolves #1820 

### How was it solved?

- [x] Add logic to verify client connection
- [x] Configure number of blocks to verify client connection

### How was it tested?
Local


services/blockchain-connector/shared/sdk/events.js
```diff
@@ -68,8 +75,19 @@ const getEventsByHeightFormatted = async (height) => {
...
 };
 
+const verifyClientConnection = async () => {
+	setInterval(() => isClientConnected = false, BLOCK_TIME * numOfBlocksReset);
+	setInterval(async () => {
+		if (!isClientConnected) {
+			logger.info('Re-instantiating API client.');
+			await instantiateClient(true);
+		}
+	}, BLOCK_TIME * numOfBlocksVerify);
```
priojeetpriyom, line 85: If we recreate a connection that is not broken, could it create any issue for ongoing node calls? Also, would it be better to reuse [getApiClient](https://github.com/LiskHQ/lisk-service/blob/51b074adc43117d70d514fc622f4ce7b8d09c402/services/blockchain-connector/shared/sdk/client.js#L85) logic, which already checks for broken connection and recreates it? Also, we do not need the `numOfBlocksReset` logic in this case.

```suggestion
	setInterval(getApiClient, BLOCK_TIME * numOfBlocksVerify);
```

services/blockchain-connector/shared/sdk/client.js
```diff
@@ -40,20 +40,18 @@ let isInstantiating = false;
 const checkIsClientAlive = () => clientCache && clientCache._channel.isAlive;
 
 // eslint-disable-next-line consistent-return
-const instantiateClient = async () => {
+const instantiateClient = async (isForceUpdate = false) => {
```
priojeetpriyom, line 43: We may not need this incase we go with the other suggestion

```suggestion
const instantiateClient = async () => {
```

services/blockchain-connector/config.js
```diff
@@ -112,4 +112,11 @@ config.job = {
 	},
 };
 
+config.verifyClientConnection = {
+	// Every n number of blocks, verify if client connection is alive
+	numOfBlocksVerify: process.env.NUM_BLOCKS_VERIFY || 15,
```
priojeetpriyom, line 117: Just a preferential suggestion, would something like `connectionVerifyBlockInterval` be more suitable naming?

services/blockchain-connector/config.js
```diff
@@ -112,4 +112,11 @@ config.job = {
...
+	// Every n number of blocks, reset client connection cache
+	numOfBlocksReset: process.env.NUM_BLOCKS_RESET || 10,
```
sameersubudhi, line 119: Why do we need two of them? Only one should be enough 🤔 
We must reset the connection only when we do not receive events for specified number of blocks (i.e. for configured numBlocks * blockTime).

services/blockchain-connector/shared/sdk/events.js
```diff
@@ -68,8 +75,19 @@ const getEventsByHeightFormatted = async (height) => {
 	return formattedEvents;
 };
 
+const verifyClientConnection = async () => {
```
sameersubudhi, line 78: I'd move all this logic to `client.js` and use Signals to communicate the receipt of events over the API client.

services/blockchain-connector/shared/sdk/index.js
```diff
@@ -125,6 +125,8 @@ const init = async () => {
 	await cacheFeeConstants();
 	await updateTokenInfo();
 
+	await verifyClientConnection();
```
sameersubudhi, line 128: This may not be necessary. The verification logic can be an IIFE within `client.js`.

sameersubudhi (changes requested):

sameersubudhi (approved):

vardan10 (approved):

PR #1848, opened 2023-09-13
Fix inconsistencies in network, market and invoke APIs

### LINKED ISSUE #1829: Fix inconsistencies in network, market and invoke APIs
### Actual behavior

1. /network/peers
    a. Total calculation is wrong when height supplied as only param and its not current height of the chain.
        http://devnet-service.liskdev.net:9901/api/v3/network/peers?height=3300
    b. ~Instead of state: any, should we accept CSV value?~
    c. ~Location not available in response~
    d. Network version regex wrong?
        http://devnet-service.liskdev.net:9901/api/v3/network/peers?networkVersion=1.0

2. /network/status
    a. network constants is assigned in below file but but does not appear in output.
        https://github.com/LiskHQ/lisk-service/blob/development/services/blockchain-indexer/shared/dataService/business/network.js#L29

4. /market/prices
    a. ~12 currency pairs are listed in config but only 6 are returned~
        https://github.com/LiskHQ/lisk-service/blob/development/services/market/config.js#L80

5. /invoke
    a. Should this have 503 incase node is not reachable?

### Expected behavior
Above mentioned inconsistencies should not exist

### Steps to reproduce
Try the above endpoints on devnet

Lisk Service v0.7.0-beta.4

### Additional Information
- All the endpoints dependendent on the Lisk application node should return 503 Service Unavailable whenever the node is unreachable

vardan10:
### What was the problem?
This PR resolves #1829 

### How was it solved?
- [x] Correct total calculation in /network/peers
- [x] Update Network version regex
- [x] Remove redundant code from /network/status
- [x] Add 503 to swagger and API documentation for /invoke

### How was it tested?
Local


services/gateway/tests/constants/generateDocs.js
```diff
@@ -584,6 +584,12 @@ const createApiDocsExpectedResponse = {
 						$ref: '#/definitions/badRequest',
 					},
 				},
+				503: {
+					description: 'Service Unavailable',
+					schema: {
+						$ref: '#/definitions/serviceUnavailable',
+					},
+				},
```
priojeetpriyom, line 592: would this api return 503 as this is only dependant on gateway?

priojeetpriyom (approved):

services/gateway/tests/constants/generateDocs.js
vardan10: Updated the code, API used to timeout incase node was not reachable. Now should throw 503.

nagdahimanshu (approved):

PR #1853, opened 2023-09-21
Add update service node page for source in Lisk Service v0.7.0 

chris529:
Closes # https://github.com/LiskHQ/lisk-docs/issues/1571

### DIFF
--- a/docs/antora/modules/ROOT/pages/update-existing-node/update-source.adoc
+++ b/docs/antora/modules/ROOT/pages/update-existing-node/update-source.adoc
@@ -1,0 +1,361 @@
+= Update PM2 commands
+Mona Bärenfänger <mona@lightcurve.io> Muhammad Talha <muhammad.talha@lightcurve.io> chris braithwaite <christopher.braithwaite@lightcurve.io>
+:description: Describes how to manage Lisk Service v0.7.0 with PM2.
+:toc:
+:idseparator: -
+:idprefix:
+:imagesdir: ../assets/images
+:source-language: bash
+// External URLs
+:url_pm2_docs: https://pm2.keymetrics.io/docs/usage/quick-start/
+:url_FLUSHALL: https://redis.io/commands/FLUSHALL
+
+On this page, you'll find the most relevant PM2 commands for running a source-based Lisk Service installation.
+For more details about PM2, see {url_pm2_docs}[PM2 docs].
+
+Moreover, we'll also talk about the steps to reset a Lisk Service node.
+
+== Various PM2 commands
+
+=== Start
+Start Lisk Service.
+
+----
+pm2 start ecosystem.config.js
+----
+
+.Response
+[%collapsible]
+====
+.Example output
+----
+[PM2] Applying action restartProcessId on app [lisk-service-gateway](ids: [ 0 ])
+[PM2] Applying action restartProcessId on app [lisk-service-blockchain-app-registry](ids: [ 1 ])
+[PM2] [lisk-service-gateway](0) ✓
+[PM2] [lisk-service-blockchain-app-registry](1) ✓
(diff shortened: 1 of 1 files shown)


chris529:
OK, just discussed, as we have now confirmed the existing PM2 commands page does not require and changes/updates for 0.7.0, I have just added in an additional link for the PM2 logs related info which points to the PM2 documentation in the existing management/source.adoc page. 👍

TalhaMaliktz:
> Then, once @TalhaMaliktz agrees to it, I'll close the PR but not merge it since we do not want to include this new page in our docs.

Hey @sameersubudhi, the idea behind these two issues: https://github.com/LiskHQ/lisk-docs/issues/1570 and https://github.com/LiskHQ/lisk-docs/issues/1571 is to target the users who already have an existing installation of either a Lisk Service Docker node or a Lisk Service Source node. 

The pages will be similar to these guides: [Update existing NPM node](https://lisk.com/documentation/lisk-core/update/npm.html) or [Update existing Binary node](https://lisk.com/documentation/lisk-core/update/binary.html). If you think that the steps to updating an existing installation of Lisk Service are very simple and the information can be relayed via just a **TIP**, which suggests checking out the latest version of the repo, then we can incorporate that in the existing docs somewhere, If not then please share the relevant steps to achieve the aforementioned task.

TalhaMaliktz:
> Hi @TalhaMaliktz, As mentioned in the comment [here](https://github.com/LiskHQ/lisk-service/pull/1853#pullrequestreview-1646391120), all users need to do is pull the latest changes and follow exactly same steps mentioned under the `Management` heading. Instead of creating a separate section the necessaty information can be added as a `TIP` or a `NOTE`. In case you still prefer to go ahead with a new section, please feel free but IMHO we're introducing unnecessary overhead.
> 
> > This information (pull the latest changes and follow the exact same steps) can be added as a note in the same page

@sameersubudhi thanks for the explanation! I agree with you! @chris529 let's add a TIP/NOTE to describe this update process. We can huddle, to discuss further!

docs/antora/modules/ROOT/pages/management/source.adoc
```diff
@@ -16,6 +16,10 @@ For more details about PM2, see {url_pm2_docs}[PM2 docs].
 
 Moreover, we'll also talk about the steps to reset a Lisk Service node.
 
+TIP: The process ID of each PM2-managed process can be seen by executing the <<list>> command. To learn more about PM2 log management please refer the {url_pm2_management}[PM2 logs documentation].
+
+NOTE: To upgrade Lisk Service to a desired version, please check out the specific version with *`git checkout <tag>`*, and follow the existing process.
```
Tschakki, line 21: What process are you referring to?

Tschakki (changes requested):
I don't see where the process of updating Lisk Service is described, it just says "follow the existing process."

docs/antora/modules/ROOT/pages/management/source.adoc
Tschakki: please note that this page does not explain any specific process. it is just a collection of commands related to managing Lisk Service. So there is no particular process below.

chris529: Yes, thats a good point. Then perhaps we could just amend the end of the note, to read as follows: 
"To upgrade Lisk Service to a desired version, please check out the specific version with *`git checkout <tag>`*, and then start Lisk Service as shown in the list of commands below.

Tschakki: Please see my comment in the other PR: There should be at least a separate section explaining the process

Tschakki (approved):
Thank you 🎉 

As mentioned in the other PR, it might be beneficial to have the guide as a separate page, but this is just a recommendation.

PR #1877, opened 2023-10-17
Fix waitForIt

### LINKED ISSUE #1869: Fix waitForIt function keeps running infinitely
### Actual behavior
The [waitForIt](https://github.com/LiskHQ/lisk-service/blob/8e8119f9575661713a458bfe0a6c69ae9aaaf20f/services/blockchain-connector/shared/utils/waitForIt.js#L20) function in connector and gateway as a bug where it keeps running indefinitely if first invocation is not successful.

### Expected behavior
- The function should exit on first successful response
- The function should not invoke the 

### Steps to reproduce
The behaviour can be reproduced with following script. Keep the script in a sibling file to the waitForIt.js in connector.

```
const delay = require("lisk-service-framework/src/delay");

let count = 0;

const waitForIt = (fn, intervalMs = 1000) => new Promise((resolve) => {
	const checkIfReady = async (that) => {
		try {
			count += 1;
			const result = await fn(count);
			clearInterval(that);
			console.log(`result: ${result}`);
			if (result !== undefined) resolve(result);
		} catch (err) {
			console.debug(`Waiting for ${intervalMs}ms ...`);
		}
	};
	const hInterval = setInterval(checkIfReady, intervalMs, this);
	checkIfReady(hInterval);
});

const checkFn = async (count) => {
	await delay(1000);
	if (count <= 1) throw new Error('custom error');
	return count;
};

const test = async () => {
	console.log('Invoking waitForIt');
	await waitForIt(checkFn);
	console.log('Done waitForIt');
};

test();
```

- v0.7.0-rc.1

sameersubudhi:
### What was the problem?
This PR resolves #1869 

### How was it solved?

- [x] Fix waitForIt
- [x] Move waitForIt to lisk-framework and publish v1.6.0
- [x] Refactor code to use waitForIt from lisk-framework

### How was it tested?
Locally with the following script

```js
const delay = (ms = 1000) => new Promise(resolve => setTimeout(resolve, ms));

let count = 0;

const waitForIt = (fn, intervalMs = 1000) => new Promise(resolve => {
	const timeout = setInterval(async () => {
		try {
			count += 1;
			const result = await fn(count);
			clearInterval(timeout);
			console.log(`result: ${result}`);
			if (result !== undefined) resolve(result);
		} catch (err) {
			console.debug(`Waiting for ${intervalMs}ms ...`);
		}
	}, intervalMs);
});

const checkFn = async count => {
	await delay(1000);
	if (count <= 1) throw new Error('custom error');
	return count;
};

const test = async () => {
	console.log('Invoking waitForIt');
	await waitForIt(checkFn);
	console.log('Done waitForIt');
};

test();
```

### DIFF
--- a/services/blockchain-connector/shared/sdk/constants/eventTopics.js
+++ b/services/blockchain-connector/shared/sdk/constants/eventTopics.js
@@ -121,7 +121,7 @@ const EVENT_TOPIC_MAPPINGS_BY_MODULE = {
 		[EVENT_NAME_CHAIN_ACCOUNT_UPDATED]: ['transactionID', 'sendingChainID'],
 		[EVENT_NAME_CCM_SENT_SUCCESS]: ['transactionID', 'sendingChainID', 'receivingChainID', 'sentCCMID'],
 		[EVENT_NAME_CCM_SENT_FAILED]: ['transactionID'],
-		[EVENT_NAME_CCM_PROCESSED]: ['transactionID', 'sendingChainID', 'receivingChainID', 'ccmID'],
+		[EVENT_NAME_CCM_PROCESSED]: ['transactionID', 'sendingChainID', 'receivingChainID'],
 		[EVENT_NAME_TERMINATED_STATE_CREATED]: ['transactionID', 'chainID'],
 		[EVENT_NAME_TERMINATED_OUTBOX_CREATED]: ['transactionID', 'chainID'],
 	},


framework/src/waitForIt.js
```diff
@@ -13,23 +13,20 @@
...
+		const timeout = setInterval(async () => {
+			try {
+				const result = await fn();
+				clearInterval(timeout);
+				resolve(result);
+			} catch (err) {
+				logger.debug(`Waiting ${intervalMs}...`);
+			}
+		}, intervalMs);
+	});
```
vardan10, line 30: 1. Should we rename this to `waitForSuccess`?
2. Since we are moving this function to framework and making it generic, should we check if result is valid? Since fn() could return improper result and not always throw.

```suggestion
const waitForSuccess = (fn, intervalMs = 1000, successValidator) => new Promise(resolve => {
		const timeout = setInterval(async () => {
			try {
				const result = await fn();
				if(!successValidator || successValidator(result)) {
					clearInterval(timeout);
					resolve(result);
				}
			} catch (err) {
				logger.debug(`Waiting ${intervalMs}...`);
			}
		}, intervalMs);
	});
```

services/blockchain-connector/shared/sdk/client.js
```diff
@@ -62,7 +62,7 @@ const instantiateClient = async (isForceUpdate = false) => {
 			return clientCache;
 		}
 
-		if ((Date.now() - instantiationBeginTime) > MAX_INSTANTIATION_WAIT_TIME) {
+		if (Date.now() - instantiationBeginTime > MAX_INSTANTIATION_WAIT_TIME) {
```
vardan10, line 65: Wont brackets offer better readability?
```suggestion
		if ((Date.now() - instantiationBeginTime) > MAX_INSTANTIATION_WAIT_TIME) {
```

vardan10 (approved):

priojeetpriyom (approved):

priojeetpriyom (approved):

PR #1878, opened 2023-10-18
Querying events endpoint with topic, senderAddress and transactionID does not work

### LINKED ISSUE #1873: Querying events endpoint with topic, senderAddress and transactionID does not work
### Actual behavior
When `events` endpoint is queried with `topic`, `senderAddress` and `transactionID` - `senderAddress` replaces `transactionID` in the internal database query.

### Expected behavior
The endpoint should work properly supporting filtering by all params simultaneously.

### Steps to reproduce
Invoke `events` endpoint above three params but provide a wrong `transactionID`

- v0.7.0-rc.1 

### Additional information
Currently, lisk service framework do not provide an api to such a query. We may need to update the knex wrapper or do a raw query.

vardan10:
### What was the problem?
This PR resolves #1873 

### How was it solved?
- [x] Fixed events endpoint to work with topic
- [x] Fixed failing integration tests
- [x] Added functional tests

### How was it tested?
Local


services/blockchain-indexer/shared/dataService/business/events.js
```diff
@@ -119,24 +120,14 @@ const getEvents = async (params) => {
...
-		};
-	}
-
 	if (params.transactionID) {
 		const { transactionID, ...remQueryParams } = queryParams;
 		queryParams = remQueryParams;
 
 		if (!params.topic) {
-			queryParams.topic = transactionID;
+			params.topic = transactionID;
```
priojeetpriyom, line 128: The reason for cloning `params` in top with `queryParams` is not to modify the user supplied params. should it be `queryParams`? if so, let's change other occurrences as well. 

```suggestion
			queryParams.topic = transactionID;
```

services/blockchain-indexer/shared/dataService/business/events.js
```diff
@@ -176,16 +167,24 @@ const getEvents = async (params) => {
...
-		rightColumn: `${eventTopicsTableSchema.tableName}.eventID`,
-	};
+	if (params.topic) {
+		const { topic } = params;
+
+		const response = await eventTopicsTable.find(
+			{
+				whereIn: { property: 'topic', values: topic.split(',') },
+				distinct: 'eventID',
+				groupBy: 'eventID',
```
priojeetpriyom, line 177: won't adding `group by` be sufficient?

services/blockchain-indexer/shared/dataService/business/events.js
```diff
@@ -101,6 +101,7 @@ const deleteEventsFromCache = async (height) => eventCache.delete(height);
 
 const getEvents = async (params) => {
 	let queryParams = _.cloneDeep(params);
+	delete queryParams.topic;
```
priojeetpriyom, line 104: Do we need to delete it?

services/blockchain-indexer/shared/dataService/business/events.js
```diff
@@ -176,16 +167,24 @@ const getEvents = async (params) => {
...
 
-	const isTopicInQuery = !!params.topic || !!queryParams.topic;
-
-	queryParams.leftOuterJoin = {
-		targetTable: eventsTableSchema.tableName,
-		leftColumn: `${eventsTableSchema.tableName}.id`,
-		rightColumn: `${eventTopicsTableSchema.tableName}.eventID`,
-	};
+	if (params.topic) {
+		const { topic } = params;
```
priojeetpriyom, line 171:
```suggestion
		const { topic, ...remQueryParams } = queryParams;
		queryParams = remQueryParams;
```

services/blockchain-indexer/shared/dataService/business/events.js
```diff
@@ -211,20 +210,8 @@ const getEvents = async (params) => {
...
-			['eventID'],
-		);
-	} else {
-		// If params doesn't contain event_topics specific column data
-		// then count all rows of event table for query optimization.
-		const { leftOuterJoin, ...remParamsWithoutJoin } = remQueryParams;
-		total = await eventsTable.count(remParamsWithoutJoin, ['id']);
-	}
+	const { order, sort, ...queryParamsWithoutOrderAndSort } = queryParams;
+	const total = await eventsTable.count({ ...queryParamsWithoutOrderAndSort, distinct: 'id' });
```
priojeetpriyom, line 214: Do we need the `distinct` here?

```suggestion
	const total = await eventsTable.count({ ...queryParamsWithoutOrderAndSort });
```

services/blockchain-indexer/shared/dataService/business/events.js
vardan10: This is because of a slight change in approach. I reverted to our old approach where we first fetch all the event IDs which include all topics specified and then query the eventsTable using those topicIDs instead of using joins.
https://github.com/LiskHQ/lisk-service/blob/e5c4d988756f51b322dc73d4f95f2234d90df1b9/services/blockchain-indexer/shared/dataService/business/events.js#L118

params do not correctly depict the way we store data in db. Since we store transactionID and senderAddress as topics, I have modified params to include all 3 cases in a comma separated format.

PR #1902, opened 2023-10-27
Lisk service performance issues

### LINKED ISSUE #1900: Lisk Service performance issues
### Actual behavior

- Indexing of blocks is very slow
- Inter-microservice RPC calls are timing out resulting in API timeouts/errors

### Expected behavior

Indexing should be faster and RPC calls

### Steps to reproduce

- Run Lisk Service commit `04d0de91e0d1af424f35d5ba63005f5a70a303f9` from the `release/0.7.0` branch against Lisk Core

`release/0.7.0` branch

sameersubudhi:
### What was the problem?

This PR resolves #1900 #1903 

### How was it solved?

- [x] Disabled blocks caching by default to improve API response times (To be investigated later)
- [x] Optimized/reduced number of calls made to the node
- [x] Tweak the block indexing logic to be slightly faster (Needs further improvement from SDK, the RPC calls to fetch the blocks take significant amount of time)
- [x] Fix event query bug with transactionID
- [x] Extend blocks response to include eventRoot
- [x] Update script names in Makefile and package.json
- [x] Update documentation as per the changes made above
- [x] Fix the inconsistency between the stakers and validators endpoint responses
- [x] Fix semgrep on CI

### How was it tested?

Ran against the testnet locally

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -21,7 +21,8 @@ jobs:
 
     container:
       # A Docker image with Semgrep installed. Do not change this.
-      image: returntocorp/semgrep
+      # TODO: Remove version pinning after https://github.com/returntocorp/semgrep/issues/9091 is resolved
+      image: returntocorp/semgrep:1.45.0
 
     # Skip any PR created by dependabot to avoid permission issues:
     if: (github.actor != 'dependabot[bot]')


services/blockchain-indexer/shared/dataService/blocks.js
```diff
@@ -84,12 +76,44 @@ const getBlocks = async (params = {}) => {
...
+		meta: {},
+	};
+
+	const response = await business.formatBlock(block);
+	blocksResponse.data.push(response);
+
+	return {
+		data: blocksResponse.data,
+		meta: {
+			count: 1,
```
vardan10, line 94: Isent this blocksResponse.data.length?
```suggestion
			count: blocksResponse.data.length,
```

services/blockchain-indexer/shared/indexer/blockchainIndex.js
```diff
@@ -512,38 +549,11 @@ const deleteIndexedBlocks = async job => {
...
-		isIndexingRunning = false;
-		logger.error(
-			`Error occurred during indexing block.\nError: ${err.message}\nStack:${err.stack}`,
-		);
-		throw new Error(err);
-	}
-
-	isIndexingRunning = false;
-};
-
```
priojeetpriyom, line 541: since we are using `upsert` operations, we could end up having mismatched data without the serialized wrapper. consider this scenario-

last height indexed is 100. now the next job will be indexing block with height 101. Both of the indexing job are executed simultaneously by the queue and we found out that the last indexed height is 100. so both will try to commit 101th block and both will succeed as this are `upsert` operation.

services/blockchain-indexer/shared/indexer/indexStatus.js
```diff
@@ -115,7 +115,6 @@ const init = async () => {
 
 	// Register event listeners
 	Signals.get('newBlock').add(checkIndexReadiness);
-	Signals.get('chainNewBlock').add(updateFinalizedHeight);
```
vardan10, line 118: Don't we need to update finalised height after every block?
This below function updates finalised height now. Does it get called after every block? 
https://github.com/LiskHQ/lisk-service/blob/development/services/blockchain-indexer/shared/messageProcessor.js#L156

nagdahimanshu: We are already updating finalised height here https://github.com/LiskHQ/lisk-service/blob/1900-lisk-service-performance-issues/services/blockchain-indexer/shared/constants.js#L195-L206

nagdahimanshu (approved):

vardan10 (approved):

priojeetpriyom (approved):

PR #1921, opened 2023-11-14
Automatic API client re-instantiation fails

### LINKED ISSUE #1920: Automatic API client re-instantiation fails
### Actual behavior

Sometimes, when the automatic API client re-instantiation fails with timeout issues, the connector gets stuck.

### Expected behavior

The API client automatic re-instantiation logic should work properly without any issues.

### Steps to reproduce

- Run Lisk Service for a couple of days against Lisk Core testnet

v0.7.0-rc.4

sameersubudhi:
### What was the problem?

This PR resolves #1920 

### How was it solved?

- [x] Fix error handling in `instantiateClient`

### How was it tested?

Locally
  - Added: `if (Math.ceil(Math.random() * 10) % 2 === 0) throw new Error('Test API client fix');` within `instantiateClient` immediately after the IPC/WS API client instantiation logic and run against Lisk Core

### DIFF
--- a/services/blockchain-connector/shared/sdk/client.js
+++ b/services/blockchain-connector/shared/sdk/client.js
@@ -38,7 +38,8 @@ let clientCache;
 let instantiationBeginTime;
 let isInstantiating = false;
 
-const checkIsClientAlive = () => clientCache && clientCache._channel.isAlive;
+const checkIsClientAlive = () =>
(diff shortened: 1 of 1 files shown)


priojeetpriyom (approved):

nagdahimanshu (approved):
LGTM 👍

PR #1926, opened 2023-11-15
Add failsafe for blockchain fork

### LINKED ISSUE #1923: Add failsafe for blockchain fork
### Actual behavior

Currently, blockchain-connector gets a `chain_deleteBlock` from node and then sends a event to blockchain-coordinator for scheduling a block deletion message. This process uses pub/sub communication before it is noted in a queue and as a result events can be lost due to various type of errors situation. Also, in this case, lisk-service will mark the outdated block(s) as finalised during subsequent indexing operations.

Example: when lisk-service is synced with the node, if connector / coordinator fails to add the block to message queue then lisk-service fails to detect the issue and old block lives in lisk service indefinately. 

### Expected behavior
Lisk service should be able to detect outdated blocks even if some micro-services fails to process blockchain fork event. Blocks should be set to finalized only when blockchain height for the block is finalised and also indexed blockID is identical to the one in blockchain.

One possible way:

Write a job which runs after every X minutes and in each run - 

1. fetches Y unfinalized indexed blocks ID (till finalized height in blockchain to avoid processing multiple times) and verifies that against the node. 
2. If there is a mismatch in the block ID, then it schedules a block deletion event for the minimum block height among the mismatched blocks. 
3. If blocks passes the verification, then it sets the blocks as finalized in DB till the max block height that has been checked.

N.B: X, Y should be configurable config values and existing block finalization logic during indexing needs to be removed.

### Steps to reproduce
1. Assume an indexed block is outdated by changing blockID of the block with height highest indexed height - 5 (or any similar number).
2. The blockID should be restored after the job execution.

- v0.7.0-rc.4

vardan10:
### What was the problem?
This PR resolves #1923 

### How was it solved?
- [x] Schedule previous block deletion incase of fork

### How was it tested?
- [ ] Local


sameersubudhi (approved):
LGTM 👍

Issue #1931, opened 2023-11-17
Add receivingChainID request param support to transactions endpoint


sameersubudhi:
### Description

Allow user to filter transactions based on `receivingChainID`.

### Acceptance Criteria

- Add support for `receivingChainID` to the `GET /transactions` endpoint
- Update swagger
- Add integration tests
- Add unit tests
- Fix unit tests for transaction fee estimates.

PR #1934, opened 2023-11-22
Add receivingChainID request param support to transactions endpoint

vardan10:
### What was the problem?
This PR resolves https://github.com/LiskHQ/lisk-service/issues/1931

### How was it solved?
- [x] Added support for receivingChainID to be currentChainID in transactions API
- [x] Fix TODOs
- [x] `/transactions` endpoint now returns pending transactions by default

### How was it tested?
Local


services/blockchain-connector/shared/sdk/schema.js
```diff
@@ -51,17 +52,16 @@ const getBlockAssetDataSchemaByModule = _module => {
 const getDataSchemaByEventName = eventName => {
```
sameersubudhi, line 52: Please update the method as discussed offline to cache the events by module and eventName instead.

services/blockchain-indexer/shared/dataService/business/pendingTransactions.js
```diff
@@ -92,7 +93,18 @@ const validateParams = async params => {
 	if (params.senderAddress) validatedParams.senderAddress = params.senderAddress;
 	if (params.recipientAddress) validatedParams.recipientAddress = params.recipientAddress;
 	if (params.moduleCommand) validatedParams.moduleCommand = params.moduleCommand;
-	if (params.receivingChainID) validatedParams.receivingChainID = params.receivingChainID;
+
+	// If recieving chainID is current chain ID then return all transactions with receivingChainID = null
```
sameersubudhi, line 97:
```suggestion
	// If receivingChainID is currentChainID then return all the transactions where receivingChainID = null
```

sameersubudhi (changes requested):

sameersubudhi (approved):
The fixes seem to work as expected, except for the issue reported in https://github.com/LiskHQ/lisk-service/issues/1937.

priojeetpriyom (approved):

PR #1948, opened 2023-11-25
Unreliable API client liveliness check

### LINKED ISSUE #1944: Unreliable API client liveliness check
### Actual behavior

The [checkIsClientAlive](https://github.com/LiskHQ/lisk-service/blob/v0.7.0-rc.5/services/blockchain-connector/shared/sdk/client.js#L41) implementation is not reliable. This sometime leads to a dead connection due to which the endpoint invocations on the node error out resulting in poor or error responses from the API endpoints.

### Expected behavior

`checkIsClientAlive` must be reliable.

### Steps to reproduce

- Query the API endpoints continuously over a period of time

v0.7.0-rc.5

vardan10:
### What was the problem?
This PR resolves #1944 

### How was it solved?
- [x] Used heartbeat to check if client is alive

### How was it tested?
- [x] Local


services/blockchain-connector/shared/sdk/client.js
```diff
@@ -32,24 +32,56 @@ const RETRY_INTERVAL = config.apiClient.instantiation.retryInterval;
...
+const checkIsClientAlive = async () =>
+	// eslint-disable-next-line consistent-return
+	new Promise(resolve => {
+		if (!clientCache || (clientCache._channel && !clientCache._channel.isAlive)) {
+			return resolve(false);
+		}
+
+		if (
+			config.isUseLiskIPCClient ||
+			Date.now() - lastClientAliveTime > CLIENT_ALIVE_ASSUMPTION_TIME
```
priojeetpriyom, line 59:
```suggestion
			Date.now() - lastClientAliveTime < CLIENT_ALIVE_ASSUMPTION_TIME
```

services/blockchain-connector/shared/sdk/client.js
```diff
@@ -32,24 +32,56 @@ const RETRY_INTERVAL = config.apiClient.instantiation.retryInterval;
...
+		const boundPongListener = () => pongListener(resolve);
+
+		clientCache._channel._ws.on('pong', boundPongListener);
+		clientCache._channel._ws.ping(() => {});
+
+		// eslint-disable-next-line consistent-return
+		setTimeout(() => {
+			clientCache._channel._ws.removeEventListener('pong', boundPongListener);
+			if (!isClientAlive) return resolve(false);
+		}, RETRY_INTERVAL);
```
priojeetpriyom, line 73: we can have this information readily available by sending a ping every 5 seconds. This would be helpful as we (almost) always invoke `checkIsClientAlive` at least once every 5 seconds timeframe.

services/blockchain-connector/shared/sdk/client.js
```diff
@@ -32,24 +32,56 @@ const RETRY_INTERVAL = config.apiClient.instantiation.retryInterval;
...
```
priojeetpriyom, line 73: Retry interval is not that much relevant. Also, has a value of 1ms by default which may not be sufficient when node has some load. Could be better to use `MAX_INSTANTIATION_WAIT_TIME` instead.

docker/example.env
```diff
@@ -30,6 +30,7 @@
 # GENESIS_BLOCK_URL='https://downloads.lisk.com/lisk/mainnet/genesis_block.json.tar.gz'
 # CLIENT_INSTANTIATION_MAX_WAIT_TIME=100,
 # CLIENT_INSTANTIATION_RETRY_INTERVAL=1,
+# CLIENT_ALIVE_ASSUMPTION_TIME=5 * 1000,
```
priojeetpriyom, line 33: If multiplication does not work here, better to set absolute value as the example.

priojeetpriyom (approved):

sameersubudhi (approved):

services/blockchain-connector/shared/sdk/client.js
sameersubudhi: It should be fine even with `0`, `MAX_INSTANTIATION_WAIT_TIME` is way too high 🙂

docker/example.env
sameersubudhi:
```suggestion
# CLIENT_ALIVE_ASSUMPTION_TIME=5000,
```

PR #1953, opened 2023-11-28
Event API takes a long time to return result when queried with topics 

### LINKED ISSUE #1952: Event API takes a long time to return result when queried with topics
### Actual behavior
Event API takes a long time to return result when queried with topics

### Expected behavior
Event API should be optimised to deliver a better performance.

### Steps to reproduce
Query this endpoint:
https://testnet-service.lisk.com/api/v3/events?offset=100&limit=100&topic=lskwtb4oevaa4jtb5q66fs4oqkac4gn9r6huehk5d

v0.7.0-rc.5

vardan10:
### What was the problem?
This PR resolves #1952 

### How was it solved?
- [x] Used joins to speed up db query
- [x] Updated gateway to only allow topics with min length 3

### How was it tested?
Local + testnet


services/blockchain-indexer/shared/dataService/business/events.js
```diff
@@ -176,42 +196,43 @@ const getEvents = async params => {
...
-				topics.push(t.slice(EVENT_TOPIC_PREFIX.TX_ID.length));
-			} else if (
-				t.startsWith(EVENT_TOPIC_PREFIX.CCM_ID) &&
-				t.length === EVENT_TOPIC_PREFIX.CCM_ID.length + LENGTH_ID
-			) {
-				// Check for CCM ID both with and without the topic prefix
-				topics.push(t.slice(EVENT_TOPIC_PREFIX.CCM_ID.length));
-			}
+		const allTopics = topic.split(',');
+		if (topic.length === LENGTH_ID) {
```
sameersubudhi, line 200: This entire if..else ladder needs to be within a for loop against the allTopics.

sameersubudhi (changes requested):
Everything else except the below looks good to me. Can we please fix this ASAP.

services/blockchain-indexer/shared/dataService/business/events.js
```diff
@@ -224,10 +248,10 @@ const getEvents = async params => {
...
 			}
 
-			const [{ id, timestamp } = {}] = await blocksTable.find({ height, limit: 1 }, [
-				'id',
-				'timestamp',
-			]);
+			if (!blocksCache[height]) {
+				blocksCache[height] = await blocksTable.find({ height, limit: 1 }, ['id', 'timestamp']);
+			}
+			const [{ id, timestamp } = {}] = blocksCache[height];
```
priojeetpriyom, line 254: Why do we need this? As height is indexed, this should be cheap operation. Also, this can produce inconsistent results when fork happens for a height.

sameersubudhi: It was added last night when we were debugging and optimizing the code. But it's also fine if we remove this caching.

priojeetpriyom (approved):

sameersubudhi (approved):
LGTM and works as expected 👍

nagdahimanshu (approved):

PR #1963, opened 2023-12-02
Fix mainnet genesis block indexing

### LINKED ISSUE #1962: Fix mainnet genesis block indexing
### Actual behavior

TODO

### Expected behavior

Indexing should progress without any issues.

### Steps to reproduce

- git pull
- git checkout 46e603c8cb0fb0b91ad06a40b394d2bce67e590e
- make build-images
- Set the following environment variables:
  - LISK_APP_WS=ws://host.docker.internal:7887
  - GENESIS_BLOCK_URL='https://lisk-mainnet-dryrun.fra1.cdn.digitaloceanspaces.com/Migrator-outputs/genesis_block.json.tar.gz'
- make up

`development` branch (commit: 46e603c8cb0fb0b91ad06a40b394d2bce67e590e)

sameersubudhi:
### What was the problem?

This PR resolves #1962 

### How was it solved?

- [x] Add safety checks to ensure that API client healthcheck is only triggered after both the nodes are synced and genesis block is downloaded
- [x] Use event-based liveness checks only for IPC clients
- [x] Update the NATS server max payload config
- [x] Update max_memory limits in pm2 config files

### How was it tested?

- [x] Locally against the mainnet
  - Follow "Steps to reproduce" from #1962 
- [x] Against the testnet, both locally and on the server

### DIFF
--- a/services/blockchain-connector/shared/sdk/client.js
+++ b/services/blockchain-connector/shared/sdk/client.js
@@ -153,14 +153,19 @@ const invokeEndpoint = async (endpoint, params = {}, numRetries = NUM_REQUEST_RE
 	} while (retries--);
 };
 
-// Check to ensure that the API Client is always alive
-const triggerRegularClientLivelinessChecks = () =>
-	setInterval(async () => {
-		const isAlive = await checkIsClientAlive();
-		if (!isAlive) instantiateClient(true).catch(() => {});
-	}, CLIENT_ALIVE_ASSUMPTION_TIME);
+// Checks to ensure that the API Client is always alive
+const resetApiClientListener = async () => instantiateClient(true).catch(() => {});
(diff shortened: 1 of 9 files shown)


services/blockchain-connector/shared/sdk/client.js
```diff
@@ -153,16 +153,14 @@ const invokeEndpoint = async (endpoint, params = {}, numRetries = NUM_REQUEST_RE
 	} while (retries--);
 };
 
-// Checks to ensure that the API Client is always alive
-const resetApiClientListener = async () => instantiateClient(true).catch(() => {});
```
ManuGowda, line 157: Please revert this changes if its needed for IPC

nagdahimanshu (approved):
LGTM 👍

PR #1986, opened 2023-12-11
Limit max web socket connection initialisation

### LINKED ISSUE #1985: Limit max web socket connection initialisation
### Description
Limit max web socket connection initialisation

### Motivation
We need to add a safety check to limit max web socket connection initialisation to 100 so connector dosen't accidentally stress out the node.

vardan10:
### What was the problem?
This PR resolves #1985 

### How was it solved?
- [x] Added limit to max web socket connection initialisation

### How was it tested?
Against Mainnet


sameersubudhi (approved):
Works as expected 👍 

Tested with pm2/docker against both mainnet and testnet on 4 core CPU and 16 GB RAM machines. The CPU usage was consistently ~80% or lower and memory usage ~55-65% (less than 10.5 GB at peak).

priojeetpriyom (approved):

PR #1994, opened 2023-12-18
Investigate lisk-core crash due to memory leak

### LINKED ISSUE #1993: Investigate lisk-core crash due to memory leak
### Actual behavior
Lisk core crashes after a while because of memory leak when connected with lisk-service as reported by devops

### Expected behavior
Lisk core should not crash because when connected with lisk-service.

- lisk-service-v0.7.2
- lisk-core-v4.0.1

priojeetpriyom:
### What was the problem?

This PR resolves #1993 

### How was it solved?

- [x] Decouple client instantiation from client fetch logic using a cached client pool
- [x] Periodically check cached clients for liveliness and replace dead clients
- [x] Remove unused config values
- [x] Add `pos_getExpectedSharedRewards` default value of `INVOKE_ALLOWED_METHODS` 

### How was it tested?
- [x] Local and cloud droplets


services/blockchain-connector/shared/sdk/client.js
```diff
@@ -186,44 +130,83 @@ const invokeEndpoint = async (endpoint, params = {}, numRetries = NUM_REQUEST_RE
...
+
+	// Initiate new clients if necessary
+	let missingClientCount = CONNECTION_LIMIT - cachedApiClients.length;
+
+	while (missingClientCount-- > 0) {
+		try {
+			await instantiateAndCacheClient();
+		} catch (err) {
+			logger.info(`Failed to instantiate new api client.\nError:${err.message}`);
 		}
```
vardan10, line 172: Should we update code to wait for clients to initialise? So that we don't end up with 0 missing clients 
```suggestion
	while (missingClientCount > 0) {
		try {
			await instantiateAndCacheClient();
			missingClientCount--;
		} catch (err) {
			logger.info(`Failed to instantiate new api client.\nError:${err.message}`);
			// TODO: Add delay here
		}
```

services/blockchain-connector/shared/sdk/client.js
```diff
@@ -158,14 +102,14 @@ const instantiateClient = async (isForceReInstantiate = false) => {
...
 	}
 };
 
 const getApiClient = async () => {
-	const apiClient = await waitForIt(instantiateClient, RETRY_INTERVAL);
-	return (await checkIsClientAlive()) ? apiClient : getApiClient();
+	if (cachedApiClients.length === 0) {
+		throw new Error(`No api client is alive!`);
+	}
+	return cachedApiClients[0];
```
vardan10, line 112: Instead should we create a variable counter which increments and returns different API clients?

```suggestion
	return cachedApiClients[counter++ % CONNECTION_LIMIT];
```

If we are always returning 1st client then maybe it dosen't make sense to create 5 different connections.

services/blockchain-connector/shared/sdk/client.js
```diff
@@ -186,44 +130,83 @@ const invokeEndpoint = async (endpoint, params = {}, numRetries = NUM_REQUEST_RE
...
+		const cacheRefreshStartTime = Date.now();
+		await refreshClientsCache();
+		logger.debug(
+			`Refreshed api client cached in ${Date.now() - cacheRefreshStartTime}ms. There are ${
+				cachedApiClients.length
+			} API client(s) in the pool.`,
+		);
+		await delay(CLIENT_ALIVE_ASSUMPTION_TIME);
+	}
+})();
```
vardan10, line 209: Better to create a molecular job than an infinite while loop?

services/blockchain-connector/shared/sdk/client.js
priojeetpriyom: It might speed up to query multiple clients simultaneously, but would also complicate the event subscription logic. I tried to prefer simplicity and reliability over perfect optimisation.

services/blockchain-connector/shared/sdk/client.js
priojeetpriyom: True, it would make code more neat. But one issue with jobs is it would use interval which does not wait for current round of checks to finish and we may endup having multiple executions of `refreshClientsCache` at same time. This would complicate the logic with locks and other checks.

nagdahimanshu (approved):
LGTM 👍

vardan10 (approved):

PR #2012, opened 2024-01-09
Optimise implementation for indexing genesis accounts balance

### LINKED ISSUE #2011: Optimise implementation for indexing genesis accounts balance
### Description

Currently, indexing genesis accounts token balance is triggered after indexing is 100% and causing connector to crash with OOM error. The implementation need to be optimized in order to avoid any crashes.

### Acceptance Criteria

- Index accounts token balance while indexing genesis assets
- Use job queue to update accounts balances

nagdahimanshu:
### What was the problem?

This PR resolves #2011 

### How was it solved?
- [x] Index accounts token balance while indexing genesis assets parallel in batches
- [x] Add error handling for
  - [x] ensureAPIClientLiveness
  - [x] New/Delete block and new transaction event in connector
  - [x] getAllPosValidators
- [x] Avoid extra calls to the node
- [x] Fix issue with event subscription

### How was it tested?
Local + Server against both networks (`mainnet` and `testnet`)

### DIFF
--- a/services/blockchain-connector/config.js
+++ b/services/blockchain-connector/config.js
@@ -127,7 +127,7 @@ config.apiClient = {
 
 // Every n milliseconds, verify if client connection is alive
 config.clientConnVerifyInterval =
-	Number(process.env.CLIENT_CONNECTION_VERIFY_INTERVAL) || 30 * 1000; // in millisecs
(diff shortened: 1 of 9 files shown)


sameersubudhi (approved):
LGTM 👍

PR #2016, opened 2024-01-25
Fix export microservice

### LINKED ISSUE #2014: Fix export microservice
### Actual behavior

- Scheduling fails if the indexing status is not exactly at 100%
- Requesting the same job over and over, schedules it multiple times
- Once a job is scheduled, the export job fails when trying to fetch transactions from the indexer

### Expected behavior

- Scheduling should work flawlessly and the above listed issues shouldn't occur

### Steps to reproduce

Repeatedly query a similar endpoint:
- `/api/v3/export/transactions?address=lsk_address&interval=date_range`

v0.7.5

sameersubudhi:
### What was the problem?

This PR resolves #2014

### How was it solved?

- [x] Add microservice dependencies for the export microservice
- [x] Allow slack in index readiness when scheduling jobs
- [x] Avoid scheduling duplicate jobs
- [x] Auto reschedule jobs in case of job timeouts
- [x] Improved logging
- [x] Refactor/Optimize code to reduce unnecessary calls to the connector/indexer from the export microservice
- [x] Add prefix handling in event topics for transactionID/ccmID
- [x] Fix old/add new unit tests

### How was it tested?

Locally and 3 server instances against the mainnet.

### DIFF
--- a/services/export/tests/unit/shared/helpers/ready.test.js
+++ b/services/export/tests/unit/shared/helpers/ready.test.js
@@ -53,10 +53,10 @@ describe('Test getIndexStatus method', () => {
 });
 
 describe('Test checkIfIndexReadyForInterval method', () => {
+	beforeEach(() => jest.resetModules());
+
 	describe('when indexing is at 100 percent', () => {
 		beforeEach(() => {
-			jest.resetModules();
-
 			const mockIndexStatus = {
 				data: {
 					genesisHeight: 23390992,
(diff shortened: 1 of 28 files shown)


services/export/tests/unit/shared/helpers/ready.test.js
```diff
@@ -0,0 +1,316 @@
...
+		const response = await getIndexStatus();
+
+		expect(response).toEqual(mockIndexStatus);
+	});
+});
+
+describe('Test checkIfIndexReadyForInterval method', () => {
+	describe('when indexing is at 100 percent', () => {
+		beforeEach(() => {
+			jest.resetModules();
```
nagdahimanshu, line 58: Not important but instead of resetting Jest modules in every tests, we can also move it outside the describe
```
beforeEach(() => jest.resetModules());
```

nagdahimanshu (approved):
Looks good and works as expected 🚀

PR #2019, opened 2024-01-29
Top balances endpoint does not consider locked amounts

### LINKED ISSUE #2018: Top balances endpoint does not consider locked balances
### Actual behavior

The `/token/balances/top` endpoint does not consider the locked balances in total amount but only the available balances.

### Expected behavior

It must consider the total amount when sorting the responses.

### Steps to reproduce

- Query the endpoint: https://service.lisk.com/api/v3/token/balances/top?tokenID=0000000000000000

v0.7.4

sameersubudhi:
### What was the problem?

This PR resolves #2018

### How was it solved?

- [x] Fixed the indexing logic for total balance to be the sum of available balance and all the locked amounts

### How was it tested?

Locally


nagdahimanshu (approved):
LGTM 👍

PR #2023, opened 2024-02-19
Can't send transactions leaving the account with less than 0.05 LSK

### LINKED ISSUE #2022: Can't send transactions leaving the account with less than 0.05 LSK
Expected behavior
User should be able to empty out an address

Actual behavior
image

Steps to reproduce
try sending a transaction that leaves the account with less than 0.05 LSK

Which version(s) does this affect? (Environment, OS, etc...)
Lisk Service 0.7.4

sameersubudhi:
### What was the problem?

This PR resolves #2022

### How was it solved?

- [x] Fix balance validation logic

### How was it tested?

Locally/ already patched on prod


nagdahimanshu (approved):
LGTM 👍

PR #2047, opened 2024-04-08
Improve indexing process

### LINKED ISSUE #2046: Improve indexing process
### Description

Currently indexing the Lisk mainchain on mainnet from scratch stresses out the node causing timeouts when invoking certain endpoints and rendering it unresponsive after a while.

### Acceptance Criteria

- Replace `token_getBalances` invocation with `token_getBalance` when updating account balances during the indexing process
- Indexing should proceed normally without causing any troubles

sameersubudhi:
### What was the problem?

This PR resolves #2046 

### How was it solved?

- [x] Replace `token_getBalances` invocation with `token_getBalance` when updating account balances during the indexing process
- [x] Include mainnet/testnet genesis blocks by default within connector microservice  

### How was it tested?

Locally, against the mainnet


nagdahimanshu (approved):
Looks good and works as expected 🚀

PR #2049, opened 2024-04-10
Token summary not updated when using HTTP API

### LINKED ISSUE #2048: Token summary not updated when using HTTP API
### Actual behavior

Token summary must be updated with every `chain_*` event.

### Expected behavior

Token summary not updated when using HTTP API after init.

### Steps to reproduce

Run Lisk Service against Lisk Core with HTTP API enabled.

v0.7.6

sameersubudhi:
### What was the problem?

This PR resolves #2048 

### How was it solved?

- [x] Update token info on blockchain state change

### How was it tested?

Locally, against the mainnet with HTTP API enabled


nagdahimanshu (approved):
LGTM 👍

PR #2065, opened 2024-05-15
Unable to fetch top token balances for certain accounts

### LINKED ISSUE #2064: Unable to fetch top token balances for certain accounts
### Actual behavior

The `/token/balances/top` endpoint doesn't return balances for all the accounts.

### Expected behavior

The `/token/balances/top` endpoint should return balances for all the existing accounts with balances.

### Steps to reproduce

Query for a migrated account (in genesis block) that contains balance but has no outgoing transactions.

v0.7.7

sameersubudhi:
### What was the problem?

This PR resolves #2064 

### How was it solved?

- [x] Fix the join query for top balance retrieval
- [x] Fix genesis account indexing count logic

### How was it tested?

Locally

### DIFF
--- a/framework/yarn.lock
+++ b/framework/yarn.lock
@@ -10,7 +10,7 @@
     "@jridgewell/gen-mapping" "^0.3.5"
     "@jridgewell/trace-mapping" "^0.3.24"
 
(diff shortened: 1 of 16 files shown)


nagdahimanshu (approved):
LGTM 👍

